Issue #4, opened 2016-01-26
TODO Before stable release


sagikazarmark:
- [x] Update email address
- [x] Update documentation link
- [x] Remove [unnecessary interface implementation](https://github.com/php-http/client-common/blob/master/src/Exception/BatchException.php#L16)
- [x] Remove composer stability config
- [x] Update composer dev alias
- [x] Update changelog

PR #5, opened 2016-01-26
Remove useless interface, minor fix in readme

joelwurtz:
See #4

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,9 +30,8 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "0.3-dev"
+            "dev-master": "1.1-dev"
         }
     },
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -30,9 +30,8 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "0.3-dev"
+            "dev-master": "1.1-dev"
```
dbu: if we did not release 1.0.0 yet, this should be 1.0-dev i think

dbu:
:+1:

composer.json
dbu: ah, just read #4 - i guess this is the same we did in the other repos, so its fine.

PR #11, opened 2016-01-31
Add flexible http client

joelwurtz:
A flexible http client implements both interface and can emulate one of the contract if not present in the injected http client.

### DIFF
--- a/spec/HttpClientFlexibleSpec.php
+++ b/spec/HttpClientFlexibleSpec.php
@@ -1,0 +1,92 @@
+<?php
+
+namespace spec\Http\Client\Common;
+
+use Http\Client\Common\HttpAsyncClientDecorator;
+use Http\Client\Common\HttpClientDecorator;
+use Http\Client\Common\HttpClientFlexible;
+use Http\Client\HttpAsyncClient;
+use Http\Client\HttpClient;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Prophet;
+
+class HttpClientFlexibleSpec extends ObjectBehavior
+{
+    function let(HttpClient $httpClient)
+    {
+        $this->beAnInstanceOf(
+            'spec\Http\Client\Common\HttpClientFlexibleStub', [
+                $httpClient
+            ]
+        );
+    }
+
+    function it_is_initializable()
+    {
+        $this->shouldHaveType('Http\Client\Common\HttpClientFlexible');
+    }
+
+    function it_is_an_http_client()
+    {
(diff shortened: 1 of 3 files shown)


src/HttpClientFlexible.php
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * @param HttpClient|HttpAsyncClient $client
+     */
+    public function __construct($client)
+    {
+        if (!($client instanceof HttpClient) && !($client instanceof HttpAsyncClient)) {
+            throw new \LogicException('Client must be an instance of Http\\Client\\HttpClient or Http\\Client\\HttpAsyncClient');
+        }
+
+        $this->httpAsyncClient = $client;
```
sagikazarmark: Just a minor logical thing: the Sync client decorator is used first, it is checked first for type. We should maintain this order in the rest of the object.

spec/HttpClientFlexibleSpec.php
```diff
@@ -0,0 +1,93 @@
...
+        $this->getClient()->shouldImplement('Http\Client\HttpClient');
+        $this->getAsyncClient()->shouldImplement('Http\Client\HttpAsyncClient');
+        $this->getClient()->shouldImplement('spec\Http\Client\Common\HttpClientFull');
+        $this->getAsyncClient()->shouldImplement('spec\Http\Client\Common\HttpClientFull');
+        $this->getClient()->shouldNotImplement('Http\Client\Common\EmulatedHttpClient');
+        $this->getAsyncClient()->shouldNotImplement('Http\Client\Common\EmulatedHttpAsyncClient');
+    }
+}
+
+class HttpClientFull implements HttpClient, HttpAsyncClient
```
sagikazarmark: This can be done using prophecy I think:

https://github.com/phpspec/prophecy#how-to-use-it

Not sure how we can access the prophet in phpspec

spec/HttpClientFlexibleSpec.php
```diff
@@ -0,0 +1,93 @@
...
+    function it_is_an_async_http_client()
+    {
+        $this->shouldImplement('Http\Client\HttpAsyncClient');
+    }
+
+    function it_throw_exception_if_invalid_client()
+    {
+        $httpClient = null;
+
+        $this->shouldThrow('\LogicException')->during('__construct', [$httpClient]);
```
sagikazarmark: You can use beConstructedWith and then duringInstantiation

sagikazarmark:
Nice :+1:

PR #13, opened 2016-01-31
Add http client router feature

### LINKED ISSUE #12: HttpClientRouter
A client accepting a map of regex -> httpclient. Given a specific request, it will check the uri of it and redirect to a specific http client implementation.

This allow to have a single pipeline for multiple library, but still allows using different client for some requests.

joelwurtz:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #12
| Documentation   | php-http/documentation#103
| License         | MIT

#### What's in this PR?

Explained in #12

#### Checklist

- [x] Documentation pull request created (if not simply a bugfix)
- [x] #11 Need to be merged before this
- [x] Matcher should be added in php-http/message (and have some basics implementations)
- [x] Squash before merge

### DIFF
--- a/src/HttpClientRouter.php
+++ b/src/HttpClientRouter.php
@@ -1,0 +1,74 @@
+<?php
+
+namespace Http\Client\Common;
+
+use Http\Client\Exception\RequestException;
+use Http\Client\HttpAsyncClient;
+use Http\Client\HttpClient;
+use Http\Message\RequestMatcher;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Route a request to a specific client in the stack based using a RequestMatcher.
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+final class HttpClientRouter implements HttpClient, HttpAsyncClient
+{
+    /**
+     * @var array
+     */
+    private $clients = [];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function sendRequest(RequestInterface $request)
+    {
+        $client = $this->chooseHttpClient($request);
+
+        return $client->sendRequest($request);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function sendAsyncRequest(RequestInterface $request)
+    {
+        $client = $this->chooseHttpClient($request);
+
+        return $client->sendAsyncRequest($request);
+    }
+
+    /**
+     * Add a client to the router.
+     *
+     * @param HttpClient|HttpAsyncClient $client
+     * @param RequestMatcher             $requestMatcher
+     */
+    public function addClient($client, RequestMatcher $requestMatcher)
+    {
+        $this->clients[] = [
+            'matcher' => $requestMatcher,
+            'client' => new FlexibleHttpClient($client),
+        ];
+    }
+
+    /**
+     * Choose an HTTP client given a specific request.
+     *
+     * @param RequestInterface $request
+     *
+     * @return HttpClient|HttpAsyncClient
+     */
+    protected function chooseHttpClient(RequestInterface $request)
+    {
+        foreach ($this->clients as $client) {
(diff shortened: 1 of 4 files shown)


src/RequestMatcher.php
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Http\Client\Common;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * RequestMatcher allow to tell if a PSR7 Request matches a specific strategy
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+interface RequestMatcher
```
dbu: did we not already add a RequestMatcher somewhere?

src/HttpClientFlexible.php
```diff
@@ -0,0 +1,39 @@
...
+use Http\Client\HttpAsyncClient;
+use Http\Client\HttpClient;
+
+/**
+ * A flexible http client, which implements both interface and will emulate
+ * one contract, the other, or none at all depending on the injected client contract.
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+class HttpClientFlexible implements HttpClient, HttpAsyncClient
```
dbu: i think this name is a bit weird. maybe HttpSyncAsyncClient ? or any better ideas?

src/RequestMatcher.php
sagikazarmark: We did, message. See https://github.com/php-http/client-common/pull/14#issuecomment-203115143

src/HttpClientFlexible.php
sagikazarmark: This is already in the package with FlexibleHttpClient. This PR needs rebase.

src/HttpClientRouter.php
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Add a client to the router.
+     *
+     * @param HttpClient|HttpAsyncClient $client
+     * @param RequestMatcher             $requestMatcher
+     */
+    public function addClient($client, RequestMatcher $requestMatcher)
+    {
+        $this->clients[] = [
```
sagikazarmark: Do you think we could find a better way to store these. Although it's an internal detail, it feels a little but weird.

sagikazarmark:
Did some improvments. Anyone review?

src/HttpClientRouter.php
dbu: having a member client and a member matcher would make more sense.

re-reading this i am confused about this code. we have a bunch of matcher, but only one single client? if none of the matchers match, we throw an exception, if any matches we always use the same client. that is not what i would expect from a router. i would have expected this to be an array of matcher, client and if a matcher matches, that client is used...

dbu: the code also indicates that this was originally the idea.

dbu: oh sorry. i misread the code. it does what i would expect. i don't see how we could store these differently. its a list of pairs [matcher, client] - i would not know how to store that differently. you could try some spl hashmap that can use objects as key if that exists, but i think its quite straightforward.

dbu:
looks good to me. can we squash the commits a bit?

also, according to the description we have no documentation yet and we should update the changelog.

src/HttpClientRouter.php
sagikazarmark: > i would have expected this to be an array of matcher, client and if a matcher matches, that client is used...

I think this is what's happening right now. Isn't it?

That one client is only used as a decorator in order to correctly emulate a client if necessary.

sagikazarmark:
Will do.

sagikazarmark:
Changelog updated, doc PR ready.

PR #14, opened 2016-03-03
Add Plugin Client

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | partially php-http/plugins#63, related php-http/plugins#64
| Documentation   | php-http/documentation#104
| License         | MIT

#### What's in this PR?

Moves Plugin client and plugins from the plugins repo

#### Checklist

- [x] Documentation pull request created (if not simply a bugfix)
- [x] Squash before merge

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Add a flexible http client providing both contract, and only emulating what's necessary
 - HTTP Client Router: route requests to underlying clients
+- Plugin client and core plugins moved here from `php-http/plugins`
 
 ### Deprecated
 
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
         "php": ">=5.4",
         "php-http/httplug": "^1.0",
         "php-http/message-factory": "^1.0",
-        "php-http/message": "^1.2"
+        "php-http/message": "^1.2",
+        "symfony/options-resolver": "^2.6|^3.0"
     },
     "require-dev": {
         "phpspec/phpspec": "^2.4",
--- a/src/Plugin/Journal.php
+++ b/src/Plugin/Journal.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Http\Client\Common\Plugin;
+
+use Http\Client\Exception;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * Records history of HTTP calls.
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+interface Journal
+{
+    /**
+     * Record a successful call.
+     *
+     * @param RequestInterface  $request  Request use to make the call
+     * @param ResponseInterface $response Response returned by the call
+     */
+    public function addSuccess(RequestInterface $request, ResponseInterface $response);
+
+    /**
+     * Record a failed call.
+     *
+     * @param RequestInterface $request   Request use to make the call
+     * @param Exception        $exception Exception returned by the call
+     */
+    public function addFailure(RequestInterface $request, Exception $exception);
+}
--- a/src/PluginClient.php
+++ b/src/PluginClient.php
@@ -1,0 +1,151 @@
+<?php
+
+namespace Http\Client\Common;
+
+use Http\Client\Common\Exception\LoopException;
+use Http\Client\Exception as HttplugException;
+use Http\Client\HttpAsyncClient;
+use Http\Client\HttpClient;
+use Http\Promise\FulfilledPromise;
+use Http\Promise\RejectedPromise;
+use Psr\Http\Message\RequestInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * The client managing plugins and providing a decorator around HTTP Clients.
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+final class PluginClient implements HttpClient, HttpAsyncClient
+{
+    /**
+     * An HTTP async client.
+     *
+     * @var HttpAsyncClient
+     */
+    private $client;
+
+    /**
+     * The plugin chain.
+     *
+     * @var Plugin[]
+     */
+    private $plugins;
+
+    /**
+     * A list of options.
+     *
+     * @var array
+     */
+    private $options;
+
+    /**
+     * @param HttpClient|HttpAsyncClient $client
+     * @param Plugin[]                   $plugins
+     * @param array                      $options {
+     *
+     *     @var int $max_restarts
+     * }
+     *
+     * @throws \RuntimeException if client is not an instance of HttpClient or HttpAsyncClient
+     */
+    public function __construct($client, array $plugins = [], array $options = [])
+    {
+        if ($client instanceof HttpAsyncClient) {
+            $this->client = $client;
+        } elseif ($client instanceof HttpClient) {
+            $this->client = new EmulatedHttpAsyncClient($client);
+        } else {
+            throw new \RuntimeException('Client must be an instance of Http\\Client\\HttpClient or Http\\Client\\HttpAsyncClient');
+        }
+
+        $this->plugins = $plugins;
+        $this->options = $this->configure($options);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function sendRequest(RequestInterface $request)
+    {
+        // If we don't have an http client, use the async call
+        if (!($this->client instanceof HttpClient)) {
+            return $this->sendAsyncRequest($request)->wait();
+        }
+
+        // Else we want to use the synchronous call of the underlying client, and not the async one in the case
+        // we have both an async and sync call
+        $pluginChain = $this->createPluginChain($this->plugins, function (RequestInterface $request) {
+            try {
+                return new FulfilledPromise($this->client->sendRequest($request));
+            } catch (HttplugException $exception) {
+                return new RejectedPromise($exception);
+            }
+        });
+
+        return $pluginChain($request)->wait();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function sendAsyncRequest(RequestInterface $request)
+    {
+        $pluginChain = $this->createPluginChain($this->plugins, function (RequestInterface $request) {
+            return $this->client->sendAsyncRequest($request);
+        });
+
+        return $pluginChain($request);
+    }
+
+    /**
+     * Configure the plugin client.
+     *
+     * @param array $options
+     *
+     * @return array
+     */
+    private function configure(array $options = [])
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'max_restarts' => 10,
+        ]);
+
+        return $resolver->resolve($options);
+    }
+
+    /**
+     * Create the plugin chain.
+     *
+     * @param Plugin[] $pluginList     A list of plugins
+     * @param callable $clientCallable Callable making the HTTP call
+     *
+     * @return callable
+     */
+    private function createPluginChain($pluginList, callable $clientCallable)
+    {
+        $firstCallable = $lastCallable = $clientCallable;
+
+        while ($plugin = array_pop($pluginList)) {
+            $lastCallable = function (RequestInterface $request) use ($plugin, $lastCallable, &$firstCallable) {
+                return $plugin->handleRequest($request, $lastCallable, $firstCallable);
+            };
+
+            $firstCallable = $lastCallable;
+        }
+
+        $firstCalls = 0;
+        $firstCallable = function (RequestInterface $request) use ($lastCallable, &$firstCalls) {
+            if ($firstCalls > $this->options['max_restarts']) {
+                throw new LoopException('Too many restarts in plugin client', $request);
+            }
+
+            ++$firstCalls;
+
+            return $lastCallable($request);
+        };
+
+        return $firstCallable;
+    }
+}
--- a/spec/FlexibleHttpClientSpec.php
+++ b/spec/FlexibleHttpClientSpec.php
@@ -79,6 +79,7 @@ class FlexibleHttpClientSpec extends ObjectBehavior
     {
         $client->implement('Http\Client\HttpClient');
         $client->implement('Http\Client\HttpAsyncClient');
+
         $client->sendRequest($syncRequest)->shouldBeCalled();
         $client->sendRequest($asyncRequest)->shouldNotBeCalled();
         $client->sendAsyncRequest($asyncRequest)->shouldBeCalled();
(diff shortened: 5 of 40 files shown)


composer.json
```diff
@@ -13,7 +13,8 @@
     "require": {
         "php": ">=5.4",
         "php-http/httplug": "^1.0",
-        "php-http/message-factory": "^1.0"
+        "php-http/message-factory": "^1.0",
+        "symfony/options-resolver": "^2.6|^3.0"
```
sagikazarmark: This is one of my concerns. Adding this dependency here. Maybe we should rethink option handling in the plugin client?

sagikazarmark:
See php-http/plugins#64

sagikazarmark:
FYI @joelwurtz

composer.json
dbu: the big benefit of options-resolver is that we are much more flexible to add things without a BC break. adding new constructor parameters can easily lead to BC break, and for these options things, its tricky to decide on the order of them - with separate parameters, the order can not be changed to make logical sense. the array is so much easier.

hm. on the other hand we have only one option for now. maybe we could instead do that manually instead of using OptionsResolver - we can add a requirement for the resolver when we add more options. for just one, its indeed overkill if its not a free include. wdyt? 

* remove the resolver dependency
* keep $options an array
* resolve manually for now
* add a comment next to the options that if we add more options, we should introduce the OptionsResolver

sagikazarmark: This is exactly what I had in my mind.

src/PluginClient.php
```diff
@@ -0,0 +1,167 @@
...
+use Http\Promise\RejectedPromise;
+use Psr\Http\Message\RequestInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * The client managing plugins and providing a decorator around HTTP Clients.
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ *
+ * TODO: make class final once removed from plugins.
```
dbu: "make class final in version 2.0, once plugins does not extend it anymore" ?

sagikazarmark: :+1:

joelwurtz:
Sorry for the delay, under high load actually :)

In fact in my mind i would also move all plugins implementations since the dependency is very strong, it does not make sense IMO to have them alone in there repository, this would add a dependency on php-http/message however,

sagikazarmark:
The reason I didn't move all plugins here is that it would add too much dependency and code here. I think having plugins separated is not bad. That's our core collection of plugins. You might want to use the plugin client, even without our plugins.

sagikazarmark:
Although I see the point in lowering the number of repositories, I feel that plugins don't really fit here. @php-http/owners any other input regarding this? It's a 1-1 ATM.

joelwurtz:
> You might want to use the plugin client, even without our plugins.

Yes there is this use case but it's only for the 1% IMO and we should not bother about that, so we should either move all code from plugins or nothing, but not something between the two.

sagikazarmark:
Well, I could even live with plugins that does not have any external dependencies. Adding any external dependencies here would make the package a monolith.

sagikazarmark:
Hm, wouldn't we add a dependency on message with php-http/client-common#13? Can we make it an optional dependency?

dbu:
i think having the clients depend on message seems really not that wrong.

i would prefer to keep everything plugin in php-http/plugin, but if we want to move it anyways we can just as well move the plugins that have no extra dependencies here, and move each plugin with special dependencies to its own repository with properly declared dependencies.

sagikazarmark:
> we want to move it anyways we can just as well move the plugins that have no extra dependencies here, and move each plugin with special dependencies to its own repository with properly declared dependencies.

Yeah, I could live with that. WDYT @joelwurtz?

IMO we had a discussion about doing this (extracting plugins with external dependencies). One possible thing we could do to avoid pluginception is that having message related, message dependent plugins in a message-plugins package. Log and cache could go into their own.

hannesvdvreken:
> move the plugins that have no extra dependencies here, and move each plugin with special dependencies to its own repository with properly declared dependencies.

Agreed!

dbu:
> message related, message dependent plugins
> in a message-plugins package. Log and cache could go into their own.

i don't like that. rather have one repo per plugin, otherwise it gets
really hard for people to figure out which plugins exist.

but maybe the concrete example is moot if we need message anyways in
client-common and can just put those plugins in here.

sagikazarmark:
> but maybe the concrete example is moot if we need message anyways in client-common and can just put those plugins in here.

Indeed. The router client is going to need it anyway.

hannesvdvreken:
> otherwise it gets really hard for people to figure out which plugins exist.

In flysystem you get local and FTP adapters out of the box (no dependencies) and it suggests (in composer.json) all of the available adapters

sagikazarmark:
Well, I think @dbu meant that we can do that, but plugins should either be completely separated or be together. However, it seems that message is going to be a dependency of this package. So plugins without dependencies and plugins depending on message could come here as well.

Pros:
- Less separate plugin packages

Cons:
- More dependency here.

dbu:
exactly, things that add no additional dependencies in this repo, everything else in a clearly named repo. afaik flysystem does the same, grouping separate plugins together makes it really confusing, its an implementors logic rather than consumer logic, and we do these things for the consumers.

sagikazarmark:
@dbu How can we proceed in the HttplugBundle if get this sorted out? Can we just drop the plugins repo dependency and use the new packages? Isn't it a BC break? Is there such thing regarding dependencies at all?

dbu:
hm. it certainly is not a patch version change to drop a dependency. but for a minor version i would be ok - it seems a bit of a grey area. but after a minor version upgrade, you should check your stuff and notice you miss the plugin. we can ease things by mentioning which plugins now need to be explicitly required if used in the changelog file.

joelwurtz:
Well if we have the php-http/message dependency we can move all plugins since other dependencies are only in suggest on the php-http/plugins package, no ?

If we look at the suggestion also (about log and cache for exemple) this is a pretty common dependency to have (at least for log ATM, and cache should follow fast looking at how it is adopted), so the suggestion will be available only for a few people.

The only plugin we can certainly move out is the Stopwatch one (and make it a dependency on the bundle or directly add him as very few people use stopwatch withtout the symfony framework)

sagikazarmark:
I would say drop suggestions and move them to separate packages.

joelwurtz:
IMO we should be more focus on the most common developer experience rather than being perfect about package separation, and even if its better to have this separated in "architecture design", i think most people wants to have the LoggerPlugin out of the box rather to have to require another package.

dbu:
i think having plugins with external dependencies in their own package is simpler to understand. its easier to say: "if you want logging: require php-http/plugin-logger" than to say: "if you want logging, there is a plugin already but you need to require psr/log".

joelwurtz:
https://twitter.com/JoelWurtz/status/715181748771033088

sagikazarmark:
Is it a cheat if I vote as well? :stuck_out_tongue:

joelwurtz:
No :)

src/Plugin/Journal.php
```diff
@@ -0,0 +1,31 @@
...
+use Http\Client\Exception;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * Records history of HTTP calls.
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+interface Journal
```
sagikazarmark: Should we move this to message instead? Are there any use cases where this might make sense?

sagikazarmark:
Plugins not (yet) moved here: Cache, Logger, Stopwatch

src/Plugin/Journal.php
dbu: or to its own repository? because its rather specific and a whole bunch
of files. and will rarely be used outside a framework integration like
symfony. moving it to message feels a bit random too, no? but journal
interfaces here in client feels odd.

sagikazarmark: The reason I suggested message is that it is mostly message related. Out of four/three dependencies three/two are messages. Implementations will probably never be in any of our repositories, because (as you said) they are mostly framework specific.

Maybe we could move the Journal together with the history plugin to it's own repository? It depends if we find any use cases for Journal apart from History plugin. Since it is relatively simple, I would say we don't necessarily need to think about any scenario.

dbu: i would move journal and history plugin to one repository and only use
the journal in the context of that plugin. that is the namespace now as
well.

sagikazarmark:
How should we proceed with the documentation? We are essentially deprecating all old plugins, but the behavior, API is mostly the same.

The namespace changed from Http\Client\Plugin to Http\Client\Common\Plugin. Should we preserve the namespace? (IMO that would make maintaining BC harder, for example throwing old exceptions)

dbu:
we can't avoid the namespace change, right? then we could just leave the old plugin package out there but deprecate it. we don't use the plugins ourselves, so people can either use the new one or the old ones, no?

regarding doc, i would just add a note that plugins initially where a separate package but deprecated in favour of those in common. no need to duplicate the documentation for plugins over that.

sagikazarmark:
> we can't avoid the namespace change, right?

Well, maybe we could, but do we want to?

> i would just add a note that plugins initially where a separate package but deprecated in favour of those in common

:+1:

dbu:
i think the namespace change makes sense. and as this is not build upon by other things, we need no tricks for BC, we can simply deprecate the old and tell people to use the new. its them instantiating all of it, so no need to extend the old class or anything.

sagikazarmark:
In that case, I would also go with #18 immediately.

dbu:
well, for the things we add here, yes!

the clients that already exist here can't be made final of course.

sagikazarmark:
> the clients that already exist here can't be made final of course.

Yep, I will add todos for them.

sagikazarmark:
Okay, what's left is to decide whether we want cache and log plugins in this repo or not.

Pros:

- Every plugin in one place
- Less packages

Cons:

- Optional dependencies
- Harder to document (https://github.com/php-http/client-common/pull/14#issuecomment-203446890)
- We cannot rely on `psr/log-implementation` and `psr/cache-implementation` virtual packages which would make dependency handling easier

I still think that we should avoid optional dependencies at all cost, because it adds a considerable maintenance overhead (which version compatible with which optional dependency). Also, I don't think that requiring to install a log-plugin package is any worse than requiring to install a log implementation.

On the other hand having everything at one place makes it easier for users to use the package (although we can document the fact of the installation).

If we can't make an agreement on this topic, I propose we have a vote about it (owners with "real" vote, community votes as feedback)

sagikazarmark:
Also let's set a deadline: this week.

ddeboer:
I’m in favour of having separate cache and log plugin packages.

We could apply Symfony’s trick, where we add one `plugins` package (with optional deps) that replaces the separate `abc-plugin` packages, just like `symfony/symfony` [replaces](https://github.com/symfony/symfony/blob/master/composer.json#L30) all `symfony/[subpackage]` packages.

Nyholm:
Sorry for not replying before. 
I can agree that we should put our plugins in the same repo as our other "tools for HTTPlug". 
 The reson **why** we do this is as far as I can understand because: 
* We should not pull unneeded dependencies. (https://github.com/php-http/plugins/pull/61)
* Getting less packages

Am I missing something?

#### About cache and log plugin:

I think (based on intuition only) that plugins that require no dependencies but Symfony OptionResolver and PSR7 should be in client-common. Other plugins should live in their own repository. This will address the issue https://github.com/php-http/plugins/pull/61

#### About the namespaces: 

The current PRs suggest that we have plugins in `Http\Client\Common\Plugin` and `Http\Client\Plugin`. Is that right? What will the new namespaces be for the plugins not moved? (logging and cache plugin)?

sagikazarmark:
> What will the new namespaces be for the plugins not moved?

I think we can use `Http\Client\Plugin` for that purpose. It makes autoloader a little bit slower, if the same namespace is used in many packages, but in production, it should be optimized anyway.

dbu:
i am +1 on having the plugins with extra dependencies in separate repositories so we can properly specify the requirements rather than have "optional dependencies".

i am -1 on the `plugins` meta package. we did that for the cmf and its quite tricky, coupling versioning of the packages for example. also, each time we add a new plugin with extra dependencies, this would lead to more things that need to get installed or in the case of virtual packages to composer errors. i would go with explicit dependencies, at least until we have a lot of separate plugins that we feel you would all want to install usually...

dbu:
Re: Namespaces: why not keep the naming as we have it in this repo and make them Http\Client\Common\Plugin? otherwise we have two different autoload roots in one package, which would be quite confusing. if we really want to avoid Common, we better move everything else to a new folder Common and change the PSR-4 root namespace accordingly. but i would prefer to put the Common in the namespace

sagikazarmark:
> coupling versioning of the packages for example.

Exactly my concern regarding the optional dependencies as well.

I don't have any arguments against using Common.

sagikazarmark:
@dbu what would happen with logging in HttplugBundle if we extract the logger plugin to a separate package?

Right now logging is enabled by default, which lets us choosing two paths:
- Require the logger plugin in the bundle as well
- Don't enable it by default

Planning to sort this out in the next few days and I am looking at other scenarios to finally decide if it could cause any problems to extract cache, log and stopwatch to separate packages.

dbu:
in the cmf, we handled this sort of "optional" dependency by saying logging defaults to `auto` but can be explicitly enabled or disabled. the setting auto means that the DI extension checks if the required class exists and if so, enables logging, otherwise disables it. i can handle the bundle side of this.

sagikazarmark:
Separate plugin repos:

https://github.com/php-http/logger-plugin
https://github.com/php-http/cache-plugin
https://github.com/php-http/stopwatch-plugin

sagikazarmark:
Documentation added, squashed, remaining plugins separated from the main repo, ready for a final review.

About single repo: for v2 version of the plugins repo we could have it as a meta package requiring all major plugins. However, we have to be careful with versioning.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Add a flexible http client providing both contract, and only emulating what's necessary
 - HTTP Client Router: route requests to underlying clients
+- Plugin client and implementations moved here from `php-http/plugins`
```
dbu: what implementations? (i assume you mean plugin implementations, but that is not clear)
should we say "Plugin client and simple plugins moved here from..."?

sagikazarmark: Good catch. We refer to these as core plugins in other places.

dbu:
are the separate repos fully ready? then i would say go ahead and merge! thanks for the effort!

CHANGELOG.md
dbu: that would do as a wording. we should also update the plugin repo
changelog to point to this move and the new separate repos.

sagikazarmark:
From my POV they are ready. Update composer dev alias, changelog and tag necessary.

PR #25, opened 2016-07-22
Add http client pool

### LINKED ISSUE #8: HttpClientPool
Having an HttpClient / HttpAsyncClient, that accept a list of differents clients and balance requests over them, should also remove a client from the list if one of the request fail.

You can see something like this here : https://github.com/elastic/elasticsearch-php/tree/master/src/Elasticsearch/ConnectionPool

joelwurtz:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #8 
| Documentation   | https://github.com/php-http/documentation/pull/134
| License         | MIT

Add a Http Client Pool as discussed in #8

### DIFF
--- a/src/HttpClientPool/LeastUsedClientPool.php
+++ b/src/HttpClientPool/LeastUsedClientPool.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace Http\Client\Common\HttpClientPool;
+
+use Http\Client\Common\Exception\HttpClientNotFoundException;
+use Http\Client\Common\HttpClientPool;
+use Http\Client\Common\HttpClientPoolItem;
+
+/**
+ * LeastUsedClientPool will choose the client with the less current request in the pool.
+ *
+ * This strategy is only useful when doing async request
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+class LeastUsedClientPool extends HttpClientPool
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function chooseHttpClient()
+    {
+        $clientPool = array_filter($this->clientPool, function (HttpClientPoolItem $clientPoolItem) {
+            return !$clientPoolItem->isDisabled();
+        });
+
+        if (0 === count($clientPool)) {
+            throw new HttpClientNotFoundException('Cannot choose a http client as there is no one present in the pool');
+        }
+
+        usort($clientPool, function (HttpClientPoolItem $clientA, HttpClientPoolItem $clientB) {
+            if ($clientA->getSendingRequestCount() === $clientB->getSendingRequestCount()) {
+                return 0;
+            }
+
+            if ($clientA->getSendingRequestCount() < $clientB->getSendingRequestCount()) {
+                return -1;
+            }
+
+            return 1;
+        });
+
+        return reset($clientPool);
+    }
+}
--- a/src/HttpClientPool/LeastUsedClientPool.php
+++ b/src/HttpClientPool/LeastUsedClientPool.php
@@ -13,12 +13,12 @@ use Http\Client\Common\HttpClientPoolItem;
  *
  * @author Joel Wurtz <joel.wurtz@gmail.com>
  */
-class LeastUsedClientPool extends HttpClientPool
+final class LeastUsedClientPool extends HttpClientPool
 {
     /**
      * {@inheritdoc}
      */
-    public function chooseHttpClient()
+    protected function chooseHttpClient()
     {
         $clientPool = array_filter($this->clientPool, function (HttpClientPoolItem $clientPoolItem) {
             return !$clientPoolItem->isDisabled();
--- a/src/HttpClientPool/RoundRobinClientPool.php
+++ b/src/HttpClientPool/RoundRobinClientPool.php
@@ -10,12 +10,12 @@ use Http\Client\Common\HttpClientPool;
  *
  * @author Joel Wurtz <joel.wurtz@gmail.com>
  */
-class RoundRobinClientPool extends HttpClientPool
+final class RoundRobinClientPool extends HttpClientPool
 {
     /**
      * {@inheritdoc}
      */
-    public function chooseHttpClient()
+    protected function chooseHttpClient()
     {
         $last = current($this->clientPool);
 
--- a/src/HttpClientPool/RoundRobinClientPool.php
+++ b/src/HttpClientPool/RoundRobinClientPool.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace Http\Client\Common\HttpClientPool;
+
+use Http\Client\Common\Exception\HttpClientNotFoundException;
+use Http\Client\Common\HttpClientPool;
+
+/**
+ * RoundRobinClientPool will choose the next client in the pool.
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+class RoundRobinClientPool extends HttpClientPool
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function chooseHttpClient()
+    {
+        $last = current($this->clientPool);
+
+        do {
+            $client = next($this->clientPool);
+
+            if (false === $client) {
+                $client = reset($this->clientPool);
+
+                if (false === $client) {
+                    throw new HttpClientNotFoundException('Cannot choose a http client as there is no one present in the pool');
+                }
+            }
+
+            // Case when there is only one and the last one has been disabled
+            if ($last === $client && $client->isDisabled()) {
+                throw new HttpClientNotFoundException('Cannot choose a http client as there is no one enabled in the pool');
+            }
+        } while ($client->isDisabled());
+
+        return $client;
+    }
+}
--- a/src/HttpClientPoolItem.php
+++ b/src/HttpClientPoolItem.php
@@ -128,7 +128,7 @@ class HttpClientPoolItem implements HttpClient, HttpAsyncClient
      *
      * @return \DateTime|null
      */
-    protected function getDisabledAt()
+    private function getDisabledAt()
     {
         return $this->disabledAt;
     }
@@ -136,7 +136,7 @@ class HttpClientPoolItem implements HttpClient, HttpAsyncClient
     /**
      * Increment the request count.
      */
-    protected function incrementRequestCount()
+    private function incrementRequestCount()
     {
         ++$this->sendingRequestCount;
     }
@@ -144,7 +144,7 @@ class HttpClientPoolItem implements HttpClient, HttpAsyncClient
     /**
      * Decrement the request count.
      */
-    protected function decrementRequestCount()
+    private function decrementRequestCount()
     {
         --$this->sendingRequestCount;
     }
@@ -152,7 +152,7 @@ class HttpClientPoolItem implements HttpClient, HttpAsyncClient
     /**
      * Enable the current client.
      */
-    protected function enable()
+    private function enable()
     {
         $this->disabledAt = null;
     }
@@ -160,7 +160,7 @@ class HttpClientPoolItem implements HttpClient, HttpAsyncClient
     /**
      * Disable the current client.
      */
-    protected function disable()
+    private function disable()
     {
         $this->disabledAt = new \DateTime('now');
     }
--- a/src/HttpClientPoolItem.php
+++ b/src/HttpClientPoolItem.php
@@ -1,0 +1,131 @@
+<?php
+
+namespace Http\Client\Common;
+
+use Http\Client\HttpAsyncClient;
+use Http\Client\HttpClient;
+use Psr\Http\Message\RequestInterface;
+use Http\Client\Exception;
+
+/**
+ * A HttpClientPoolItem represent a HttpClient inside a Pool.
+ *
+ * It is disabled when a request failed and can be reenable after a certain number of seconds
+ * It also keep tracks of the current number of request the client is currently sending (only usable for async method)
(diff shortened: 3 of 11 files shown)


src/HttpClientPoolItem.php
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        ++$this->sendingRequestCount;
+
+        return parent::sendAsyncRequest($request)->then(function ($response) {
+            --$this->sendingRequestCount;
+
+            return $response;
+        }, function ($exception) {
+            $this->disable();
```
dbu: not sure if this is a good idea. it could mean that flaky backends would lead to random client objects being disabled over time.

dbu: ah, found the reenable. but still, is your assumption that one client stops working while others continue to work? as this is not multithread or multiprocess code, i don't expect this to be a relevant use case.

or is the idea to use a bunch of clients to do round robin on the client to different instances of the same server application rather than on a load balancer in front of the servers?

dbu:
the implementation looks sane enough to me. i am not sure if one is supposed to do this - i thought with ES the idea is you talk to the cluster and ES does heal itself and make sure one of its nodes is responding. if you are writing, this could even create a huge mess if the ES node runs into a split brain - you would then write to both halves...

src/HttpClientPoolItem.php
```diff
@@ -0,0 +1,127 @@
...
+ * A HttpClientPoolItem represent a HttpClient inside a Pool.
+ *
+ * It is disabled when a request failed and can be reenable after a certain number of seconds
+ * It also keep tracks of the current number of request the client is currently sending (only usable for async method)
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+class HttpClientPoolItem extends FlexibleHttpClient
+{
+    /** @var int Number of request this client is actually sending */
```
Nyholm: s/actually/currently/ ?

src/Exception/NotFoundHttpClientException.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Http\Client\Common\Exception;
+
+use Http\Client\Exception\TransferException;
+
+/**
+ * Thrown when a http client cannot be chose in a pool.
```
sagikazarmark: s/chose/chosen

src/Exception/NotFoundHttpClientException.php
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+class NotFoundHttpClientException extends TransferException
```
sagikazarmark: HttpClientNotFound?

src/HttpClientPoolItem.php
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * A HttpClientPoolItem represent a HttpClient inside a Pool.
+ *
+ * It is disabled when a request failed and can be reenable after a certain number of seconds
+ * It also keep tracks of the current number of request the client is currently sending (only usable for async method)
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+class HttpClientPoolItem extends FlexibleHttpClient
```
sagikazarmark: Is this actually necessary? Can't we just wrap the injected client into a FlexibleHttpClient and leave it final?

src/HttpClientPoolItem.php
```diff
@@ -0,0 +1,127 @@
...
+        $this->reenableAfter = $reenableAfter;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function sendRequest(RequestInterface $request)
+    {
+        if ($this->isDisabled()) {
+            throw new Exception\RequestException('Cannot send request as this client as been disabled', $request);
```
sagikazarmark: s/as/has

src/Exception/NotFoundHttpClientException.php
sagikazarmark: Or NoHttpClientFound?

src/HttpClientPool/LeastUsedClientPool.php
```diff
@@ -0,0 +1,49 @@
...
+class LeastUsedClientPool extends HttpClientPool
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function chooseHttpClient()
+    {
+        $clientPool = [];
+
+        foreach ($this->clientPool as $clientPoolItem) {
```
sagikazarmark: Wouldn't an array filter be cleaner?

src/HttpClientPoolItem.php
joelwurtz: > or is the idea to use a bunch of clients to do round robin on the client to different instances of the same server application rather than on a load balancer in front of the servers?

Yes

Having this can remove SPOF from application, as every request / frontend / command act as load balancer

In fact this is something which his already existing in the elasticsearch-php, as my goal is to propose a solution to integrate httplug into their lib this will allow them to completly remove their code about http client and only relying on us.

Also i think this make sense as elasticsearch is not a exclusive solution using cluster mecanism.

joelwurtz:
> the implementation looks sane enough to me. i am not sure if one is supposed to do this - i thought with ES the idea is you talk to the cluster and ES does heal itself and make sure one of its nodes is responding. if you are writing, this could even create a huge mess if the ES node runs into a split brain - you would then write to both halves...

It depens on your configuration, in our projects we mainly set all nodes to be master with data to have a real cluster and also always make sure that if a node fail the system is still responding.

Having one master fail to this principle, as if the master node is down the entire application is down, it's the same if we put a load balancer in front of elasticsearch as the SPOF will be on the loadbalancer. By directly load balancing into the code we are sure that they will never be a SPOF between our application and the elasticsearch cluster.

src/HttpClientPoolItem.php
joelwurtz: No, as there is the disable / enable mecanism

src/HttpClientPool/LeastUsedClientPool.php
joelwurtz: :+1:

src/HttpClientPoolItem.php
sagikazarmark: Why not? It can be maintained it in this client:

``` php
<?php

final class HttpClientPoolItem
{
    private $client;

    public function __construct($client, $reenableAfter = null)
    {
        $this->client = new FlexibleHttpClient($client);

        $this->reenableAfter = $reenableAfter;
    }
}
```

joelwurtz: Oh, you mean like that, i think this is pretty much the same, however there is less inheritance in this case, so :+1:

src/HttpClientPool/LeastUsedClientPool.php
```diff
@@ -0,0 +1,45 @@
...
+     * {@inheritdoc}
+     */
+    public function chooseHttpClient()
+    {
+        $clientPool = array_filter($this->clientPool, function (HttpClientPoolItem $clientPoolItem) {
+            return !$clientPoolItem->isDisabled();
+        });
+
+        if (0 === count($clientPool)) {
+            throw new HttpClientNotFoundException('Cannot choose a http client as there is no one present in the pool');
```
dbu: should we just force-use a disabled client in this case rather than do a hard failure? it would reduce the time we have problems if the backend went away for a short time.

src/HttpClientPool/RoundRobinClientPool.php
```diff
@@ -0,0 +1,41 @@
...
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+class RoundRobinClientPool extends HttpClientPool
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function chooseHttpClient()
+    {
+        $last = current($this->clientPool);
```
dbu: is there a way how we could randomize with which client to start, maybe in the constructor? that would result in better load balancing of web requests, if we only do one request with each client.

src/HttpClientPoolItem.php
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * A HttpClientPoolItem represent a HttpClient inside a Pool.
+ *
+ * It is disabled when a request failed and can be reenable after a certain number of seconds
+ * It also keep tracks of the current number of request the client is currently sending (only usable for async method)
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+class HttpClientPoolItem implements HttpClient, HttpAsyncClient
```
sagikazarmark: Shouldn't this be final?

src/HttpClientPool/RoundRobinClientPool.php
sagikazarmark: Isn't the least used client exactly for that?

src/HttpClientPoolItem.php
dbu: i could see the use for customizing the pool item behaviour. disable is protected, indicating this is meant to be extended for custom use cases.

src/HttpClientPool/RoundRobinClientPool.php
dbu: both are only within the same php process. when using a client in web requests, there hopefully is only one request to some other service, not several, and in that case this only makes sense if we randomize

joelwurtz: We can maybe add a third strategy which randomly choose a client between the one that are available ?

src/HttpClientPoolItem.php
joelwurtz: > i could see the use for customizing the pool item behaviour. disable is protected, indicating this is meant to be extended for custom use cases.

Yes, i don't want to add that here, but i can see use case where you want to persist metadata (disable, counter of request) of the HttpClientPoolItem, that's why i want to allow inheritance of this (so we can add, in the futur, an extended library using a storage to share this metadata across php process)

src/HttpClientPool/RoundRobinClientPool.php
dbu: that would make sense i think, and should be easy: shuffle the pool before doing what we do here. we might even extend this class and do shuffle, then call parent.

joelwurtz: Done see `RandomClientPool` also added in the PR docs

src/HttpClientPoolItem.php
joelwurtz: I add some methods in the class to make it better extendable given this view, tell me WDYT

src/HttpClientPoolItem.php
```diff
@@ -53,12 +50,12 @@ public function sendRequest(RequestInterface $request)
...
         try {
-            ++$this->sendingRequestCount;
+            $this->incrementRequestCount();
             $response = $this->client->sendRequest($request);
-            --$this->sendingRequestCount;
+            $this->decrementRequestCount();
         } catch (Exception $e) {
             $this->disable();
-            --$this->sendingRequestCount;
+            $this->decrementRequestCount();
```
dbu: if we where using php 5.5 only, we could use finally for this. but we are on 5.4 it seems :-(

src/HttpClientPoolItem.php
dbu: the class is now exposing many protected methods. this limits flexibility when refactoring. maybe we should keep that all private for now and refactor when you build that persistent solution. you might realize you need a different implementation then - and if they are now private, we can still change everything we need.

probably we need to inject either the strategy or the storage (in-memory, memcache, whatnot) to avoid needing classes for each permutation in-memory+round robin, in-memory+least used, memcache+round robin, ... but i would not work on that yet and just keep things as simple and internal as possible so we can extend later.

src/HttpClientPool/LeastUsedClientPool.php
```diff
@@ -0,0 +1,45 @@
...
+ * This strategy is only useful when doing async request
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+class LeastUsedClientPool extends HttpClientPool
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function chooseHttpClient()
```
Nyholm: All the chooseHttpClient should be protected, right?

src/HttpClientPoolItem.php
```diff
@@ -0,0 +1,167 @@
...
+    {
+        return $this->sendingRequestCount;
+    }
+
+    /**
+     * Return when this client has been disabled or null if it's enabled.
+     *
+     * @return \DateTime|null
+     */
+    protected function getDisabledAt()
```
dbu: can we make all of these methods private for now? we should refactor when we actually implement something with persistent pool information. then we might need to adjust things anyways, which will be easier if these are private for now. maybe even make this class final for now, so that we can refactor without any worry about BC (apart from the public methods and constructor)

src/HttpClientPool/LeastUsedClientPool.php
joelwurtz: This can be done by setting 0 on the http client pool item

src/HttpClientPoolItem.php
joelwurtz: done

dbu:
can you add an entry to the changelog please?

apart from that looks ready to me.

sagikazarmark:
Sorry for not being active in this one. I am happy to merge when there is a change log entry.

PR #27, opened 2016-07-26
adjust phpdoc of error plugin

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

#### What's in this PR?

Phpdoc adjust.

#### Why?

The doc said "by default" but there is no options to change behaviour.


dbu:
lowest version tests fail on master as well. we should investigate why that happens, but could still merge this if you agree.

Nyholm:
Thank you for this PR.

PR #30, opened 2016-07-26
Fix for the build with --prefer-lowest

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | n/a
| License         | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     "require": {
         "php": ">=5.4",
         "php-http/httplug": "^1.0",
-        "php-http/message-factory": "^1.0",
+        "php-http/message-factory": "^1.0.2",
         "php-http/message": "^1.2",
         "symfony/options-resolver": "^2.6|^3.0"
     },


sagikazarmark:
Do we actually need that version here? Or just a dependency needs it? I would rather find the reason why this doesn't work as intended instead of this update.

Nyholm:
I know you would Mark. This is me being pragmatic to solve the build error. The difference between message-factory:1.0.0 and 1.0.2 is that the latter has Puli bindings (if I remember correctly)

So no, we do not actually *need* that version here.

dbu:
not sure if its worth the effort to figure out why it does not work. there is no harm in upping the requriement to the factory...

sagikazarmark:
I found the problem: https://github.com/composer/composer/issues/5355

PR #33, opened 2016-07-30
Added missing import

GrahamCampbell:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

#### What's in this PR?

Adds a missing import to fix fatal errors.

#### Why?

image

#### Checklist

#### To Do

- [ ] If the PR is not complete but you want to discuss the approach, list what remains to be done here

### DIFF
--- a/src/Plugin/HistoryPlugin.php
+++ b/src/Plugin/HistoryPlugin.php
@@ -2,6 +2,7 @@
 
 namespace Http\Client\Common\Plugin;
 
+use Exception;
 use Http\Client\Common\Plugin;
 use Http\Client\Exception;
 use Psr\Http\Message\RequestInterface;
--- a/src/Plugin/HistoryPlugin.php
+++ b/src/Plugin/HistoryPlugin.php
@@ -2,7 +2,6 @@
 
 namespace Http\Client\Common\Plugin;
 
-use Exception;
 use Http\Client\Common\Plugin;
 use Http\Client\Exception;
 use Psr\Http\Message\RequestInterface;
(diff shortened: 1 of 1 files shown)


src/Plugin/HistoryPlugin.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Http\Client\Common\Plugin;
 
+use Exception;
```
sagikazarmark: Conflicts with the other exception import. Also, our convention is not to import root namespaces classes, but use them directly.

sagikazarmark:
This might not be that easy though. The Journal only accepts Httplug exceptions, so we need to check the exception type in the rejection handler.

@Nyholm you created this part AFAIK. What was the convention? The Journal only collects HTTP related exceptions, right? (So we can show them in the debug toolbar?) Any other exception is out of scope IMO.

sagikazarmark:
Thanks @GrahamCampbell. So Styleci uses HTTPlug now? 😄 Exciting. Our code is used to test our code?

GrahamCampbell:
Hi, yeh, the link is https://github.com/KnpLabs/php-github-api/issues/409. I was testing the new library on StyleCI (now in production). If we can't get this resolved soonish, I will need to rollback the deployment.

GrahamCampbell:
How does this look now, I've taken off the typehint so we can deal with any eventuality, and added an instanceof check.

GrahamCampbell:
Hmm, maybe we don't actually need to change anything here. Looks like the other library is actually broken here:

image

Throwing another exception there should not be allowed, right?

GrahamCampbell:
Closing this. There is no bug in your library. The other library is just misusing it, right?

GrahamCampbell:
How does https://github.com/KnpLabs/php-github-api/pull/410 look to you please?

---

Thanks for the cool library btw. :)

PR #35, opened 2016-07-31
Make sure we catch all exceptions

### LINKED ISSUE #34: Clear exception usage in Plugins
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 1.1-

#### Actual Behavior

Plugin implementations only accept `Http\Client\Exception` instances in their rejection handler ([here](https://github.com/php-http/client-common/blob/master/src/Plugin/HistoryPlugin.php#L43) and [here](https://github.com/php-http/client-common/blob/master/src/Plugin/RetryPlugin.php#L65)) while there is no such limitation in the promise spec or the [RejectedPromise](https://github.com/php-http/promise/blob/master/src/RejectedPromise.php#L20).

#### Expected Behavior

Plugins should accept any kind of exceptions and handle only which are acceptable for them and rethrow the rest. This is not a breaking change since we just need to widen the accepted exception types in the plugins.

#### Steps to Reproduce

See https://github.com/php-http/client-common/pull/33

#### Possible Solutions
As said, widen the range of accepted exception types.

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | ~~yes~~ ~~no~~ yes
| Deprecations?   | no
| Related tickets | fixes #34, related to #33 
| Documentation   | 
| License         | MIT

#### What's in this PR?

I widening the type hints to catch all exceptions.

### DIFF
--- a/src/HttpAsyncClientEmulator.php
+++ b/src/HttpAsyncClientEmulator.php
@@ -2,7 +2,6 @@
 
 namespace Http\Client\Common;
 
-use Http\Client\Exception;
 use Http\Promise;
 use Psr\Http\Message\RequestInterface;
 
@@ -29,7 +28,7 @@ trait HttpAsyncClientEmulator
     {
         try {
             return new Promise\FulfilledPromise($this->sendRequest($request));
-        } catch (Exception $e) {
+        } catch (\Exception $e) {
             return new Promise\RejectedPromise($e);
         }
     }
--- a/src/Plugin/HistoryPlugin.php
+++ b/src/Plugin/HistoryPlugin.php
@@ -40,8 +40,10 @@ final class HistoryPlugin implements Plugin
             $journal->addSuccess($request, $response);
 
             return $response;
-        }, function (Exception $exception) use ($request, $journal) {
-            $journal->addFailure($request, $exception);
+        }, function (\Exception $exception) use ($request, $journal) {
+            if ($exception instanceof Exception) {
+                $journal->addFailure($request, $exception);
+            }
 
             throw $exception;
         });
--- a/src/PluginClient.php
+++ b/src/PluginClient.php
@@ -3,7 +3,6 @@
 namespace Http\Client\Common;
 
 use Http\Client\Common\Exception\LoopException;
-use Http\Client\Exception as HttplugException;
 use Http\Client\HttpAsyncClient;
 use Http\Client\HttpClient;
 use Http\Promise\FulfilledPromise;
@@ -79,7 +78,7 @@ final class PluginClient implements HttpClient, HttpAsyncClient
         $pluginChain = $this->createPluginChain($this->plugins, function (RequestInterface $request) {
             try {
                 return new FulfilledPromise($this->client->sendRequest($request));
-            } catch (HttplugException $exception) {
+            } catch (\Exception $exception) {
                 return new RejectedPromise($exception);
             }
         });
--- a/src/Plugin/RetryPlugin.php
+++ b/src/Plugin/RetryPlugin.php
@@ -3,7 +3,6 @@
 namespace Http\Client\Common\Plugin;
 
 use Http\Client\Common\Plugin;
-use Http\Client\Exception;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
@@ -62,7 +61,7 @@ final class RetryPlugin implements Plugin
             }
 
             return $response;
-        }, function (Exception $exception) use ($request, $next, $first, $chainIdentifier) {
+        }, function (\Exception $exception) use ($request, $next, $first, $chainIdentifier) {
             if (!array_key_exists($chainIdentifier, $this->retryStorage)) {
                 $this->retryStorage[$chainIdentifier] = 0;
             }


src/Plugin/Journal.php
```diff
@@ -25,7 +24,7 @@ public function addSuccess(RequestInterface $request, ResponseInterface $respons
      * Record a failed call.
      *
      * @param RequestInterface $request   Request use to make the call
-     * @param Exception        $exception Exception returned by the call
+     * @param \Exception        $exception Exception returned by the call
      */
-    public function addFailure(RequestInterface $request, Exception $exception);
+    public function addFailure(RequestInterface $request, \Exception $exception);
```
Nyholm: Here is the BC break. Possible solution to avoid this is to introduce a new Journal interface. Could we do something better?

src/HttpAsyncClientEmulator.php
```diff
@@ -29,7 +28,7 @@ public function sendAsyncRequest(RequestInterface $request)
     {
         try {
             return new Promise\FulfilledPromise($this->sendRequest($request));
-        } catch (Exception $e) {
+        } catch (\Exception $e) {
```
GrahamCampbell: What about throwables?

Nyholm: The RejectedPromise() takes a \Exception in its constructor. That's why we should use \Exception here.

GrahamCampbell: What if we modified that to take a throwable?

GrahamCampbell:
👍 for merging this, then look at getting throwables to work too after?

sagikazarmark:
Is there a generally accepted way for supporting both PHP 5.x and 7.x? Remove typehint or create a fallback interface, etc?

src/Plugin/Journal.php
```diff
@@ -25,7 +24,7 @@ public function addSuccess(RequestInterface $request, ResponseInterface $respons
      * Record a failed call.
      *
      * @param RequestInterface $request   Request use to make the call
-     * @param Exception        $exception Exception returned by the call
+     * @param \Exception       $exception Exception returned by the call
      */
-    public function addFailure(RequestInterface $request, Exception $exception);
+    public function addFailure(RequestInterface $request, \Exception $exception);
```
sagikazarmark: This is a BC break, you cannot do that. All implementations of this interface will be broken.

src/Plugin/HistoryPlugin.php
```diff
@@ -40,7 +39,7 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
             $journal->addSuccess($request, $response);
 
             return $response;
-        }, function (Exception $exception) use ($request, $journal) {
+        }, function (\Exception $exception) use ($request, $journal) {
```
sagikazarmark: You need to add a typecheck here and only add the exception to the journal if it is an HTTP exception.

src/Plugin/Journal.php
Nyholm: Here is the BC break. Possible solution to avoid this is to introduce a new Journal interface. Could we do something better?

src/Plugin/HistoryPlugin.php
Nyholm: Are we happy with the Journal only accepting HTTPlug exceptions? Or should we try to find an other solution?

src/Plugin/HistoryPlugin.php
```diff
@@ -40,8 +40,10 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
             $journal->addSuccess($request, $response);
 
             return $response;
-        }, function (Exception $exception) use ($request, $journal) {
-            $journal->addFailure($request, $exception);
+        }, function (\Exception $exception) use ($request, $journal) {
+            if ($exception instanceof Exception) {
+                $journal->addFailure($request, $exception);
```
GrahamCampbell: How can we see other exceptions then?

GrahamCampbell:
> Is there a generally accepted way for supporting both PHP 5.x and 7.x? Remove typehint or create a fallback interface, etc?

Removing the typehint, or turning the throwable into an actual exception like symfony and laravel do.

src/Plugin/HistoryPlugin.php
Nyholm: I guess we can't. Or we shouldn't. The journal is only meant to HTTP related exceptions.

GrahamCampbell:
Odd that the StyleCI PR analysis is failing.

sagikazarmark:
I see strange SSH key added emails coming from StyleCI and that was certainly not me.

GrahamCampbell:
Looking at the logs, that's because GitHub is giving us `Permission denied (publickey). fatal: Could not read from remote repository. Please make sure you have the correct access rights and the repository exists.` when we try to analyze the PR, so the system tries to repair the deploy key.

src/Plugin/HistoryPlugin.php
sagikazarmark: Other exceptions should be handled by your code or logged/"handled" by the framework (end up in 500 usually). This Journal is to record HTTP messages and potential failures of them. This is the same questions as domain/http related exceptions.

GrahamCampbell: Ok, makes sense. :)

dbu:
so, ready to merge? @Nyholm can you try to squash the commits and see if we get styleci to also analyze "pr" and not only "push"? tried to restart the build and it said its all good, but i guess that is the "push" and not the "pr". no idea how to make it rebuild "pr"

Nyholm:
I think the PR is ready to be merged. 

The PR is squashed and rebased.

src/PluginClient.php
```diff
@@ -79,7 +78,7 @@ public function sendRequest(RequestInterface $request)
         $pluginChain = $this->createPluginChain($this->plugins, function (RequestInterface $request) {
             try {
                 return new FulfilledPromise($this->client->sendRequest($request));
-            } catch (HttplugException $exception) {
+            } catch (\Exception $exception) {
```
dbu: this is a BC break for custom plugins, right? before, their failure callback would accept only Http\Client\Exception, but now it could be called with \Exception. this will lead to a fatal error with parameter mismatch, where before it would be an exception that broke the chain of promises...

dbu: i do think its correct to do this, but maybe we need a BC layer for this. maybe an option to the PluginClient for this new behaviour, that is off by default. version 2 can then drop the option and always catch all exceptions.

GrahamCampbell: i don't think this is breaking because the rejected promise class always said that it could take things that weren't http plug exceptions

GrahamCampbell:
👍  for merging

src/PluginClient.php
Nyholm: If one have written an function HttplugException (just like we did in HistoryPlugin) it will fail with this update. But as @GrahamCampbell says, the rejected promise class always said it would be an \Exception. 

So: This is a bugfix, others may also suffer from this bug. But this is no BC break.

cursedcoder:
👍

src/PluginClient.php
dbu: hm, okay. but if our own classes did not adhere to what we said, we should make sure this will go into a minor version change and that the changelog is clear about it.

sagikazarmark: So we actually have to do this? AFAIK the promise does no limit you what exceptions you use, bet as we agreed, we should only handle HTTPlug exceptions, others should be rethrown ASAP.

sagikazarmark: Actually I agree with @dbu: the differing catch IS a BC break.

Nyholm: What about the RetryPlugin for example? It should not retry if it catches a non-httplug exception?

Also, If I would like to write my own History plugin to log all exceptions I cant do that?

The reson I do not think it is a BC break is because the return value of Plugin::handleRequest is a Promise. And a Promise can throw an \Exception. 

> So plugins need to expect (in typehints) that they might receive any kind of exceptions, but we only want to handle HTTPlug ones on purpose.

I assume by "we only want ..." you mean our plugins? Not all custom plugins as well?

joelwurtz: > The reson I do not think it is a BC break is because the return value of Plugin::handleRequest is a Promise. And a Promise can throw an \Exception.

It's not hard coded, but it's specified in the docblock of the HttpAsyncClient : https://github.com/php-http/httplug/blob/master/src/HttpAsyncClient.php#L22

So for me handling `\Exception` here is a BC break according to the docblock, i made this PR https://github.com/php-http/httplug/pull/116 to have this on the code on not only the docblock

And like i said in https://github.com/php-http/client-common/issues/34 i'm not against this change however we **should** be consistent and also deal with the case where the resolved value of our promise is not a `ResponseInterface` if we agree to do this.

Nyholm: > It's not hard coded, but it's specified in the docblock of the HttpAsyncClient : https://github.com/php-http/httplug/blob/master/src/HttpAsyncClient.php#L22

I missed that. That changes things. I agree, this is a BC break.

sagikazarmark: This also indicates a BC break: https://github.com/php-http/httplug/blob/master/src/HttpAsyncClient.php#L24

Meaning if the exception is not an HTTP one, it should be thrown immediately.

So to sum up: the Promise actually allows to receive a `\Exception` instance in the rejection callback, however Promises in the HTTP flow should only receive HTTP related exceptions, everything else should immediately be thrown.

Nyholm: How do we move forward? How do we avoid issues like this https://github.com/php-http/client-common/pull/33 to occur?

We have to make sure that no plugin ever receives anything else than a HTTPlug exception.

joelwurtz: > the Promise actually allows to receive a \Exception instance in the rejection callback, however Promises in the HTTP flow should only receive HTTP related exceptions, everything else should immediately be thrown.

That's why we may need to add an HttpPromise interface, and use this one, instead of the Promise, which enforce this and offer better visibility for devs when creating plugins and client.

joelwurtz: > How do we move forward? How do we avoid issues like this #33 to occur?

We need to take a decision of one of this three choices : 

 1. Do not change plugins / promise / httplug and add better documentation on the context of a promise in httplug
 2. Enforce this with a new interface and class like proposed in https://github.com/php-http/httplug/pull/116 (IMO, this is not a BC break but that's discutable)
 3. Make a BC break and allows everything as a rejected or resolved value and add checks everywhere where we use promise

In my opinion, 2 is the best solution as even if we add documentation, there are still people who doesn't read it and it would lead to the same error

However i'm ok with 1

3 is the worst thing to do IMO as we will increase all our libs with uneeded complexity.

dbu: what situations do we have non-http exceptions being thrown around? is that an expected behaviour that the application should handle, or is it only on configuration errors and the like? on misconfiguration, the plugins should not need to act, the developer will have to fix the bug he or she created. not going through the plugin chain in such cases will make debugging easier actually.

GrahamCampbell: Whenever we want to throw a domain specific exception.

GrahamCampbell: https://github.com/KnpLabs/php-github-api/blob/master/lib/Github/HttpClient/Plugin/GithubExceptionThrower.php

sagikazarmark: > Make a BC break and allows everything as a rejected or resolved value and add checks everywhere where we use promise

But allowing everything from the promise point of view is what we have at the moment. However, during the HTTP request flow, the promise should only receive HTTP exceptions IMO. Non-HTTP exceptions (either emitted by plugins or not) should not be caught in the first place.

The other thing is that domain related exceptions should probably not be thrown during the HTTP flow, but after it is finished in the API Client class or whatever consumes the HTTP endpoint. Even if they should be done for all requests (eg. rate limiting).

May this be the root of the problem? https://github.com/php-http/guzzle6-adapter/blob/master/src/Promise.php#L65

The phpdoc says the stored exception should be an HttplugException, but we store runtime ones there too?

About HttpPromise: 

1. It is BC break (EVERY interface change is BC break)
2. Extending the Promise interface doesn't help as there is no return type overloading (even if it is only typehint, but think about it as PHP 7 return typehints are implemented)

What I would do: document better and find why a non-http exception is not thrown immediately.

sagikazarmark: Hm, every exception is caught here: https://github.com/php-http/promise/blob/master/src/RejectedPromise.php#L36

@joelwurtz would your PR to HTTPlug solve this issue?

sagikazarmark: Maybe there is a fourth option: add an Exception interface to Promise and accept only those in promises, every other exception needs to be thrown by definition. Then in plugins we still need to check whether the received exception is an Http one (Http ones should be Promise ones too). The difference is that this way in the promise handling we can force a type of exception without forcing `\Exception` (could be a solution for throwables too). It is a BC break from Promise point of view (exception typehints changed), however Promise is kind of an internal thing from httplug point of view, so I have no problem with tagging v2.

dbu: the promise interface would mean that HTTPlug starts to depend on promise, to implement that interface on its exceptions. what is the state of promise PSR?

i think we should not handle domain exception in HTTPlug plugins. the plugin chain is for handling HTTP, not for handling domain exceptions. the github api exceptions don't transport the response object - which is correct, but makes it really pointless to continue handling the exception in http plugins. PluginClient should not be an arbitrary stack of promises. if a user wants to build a stack of domain promises, they should not use http promises for that, imho.

Nyholm: So: write ducumentation that explains that only HTTPlug exceptions can be thrown in the plugin. And that's it? 

What if a plugin cause a runtime exception? How is that handled? Should the next plugin see it?

GrahamCampbell: Mergy mergy! :trollface:

sagikazarmark: The problem is that the used RejectedPromise accepts any kind of exceptions, so if an exception is thrown, the rejected promise will call the reject callback with any exception. IMO there are two/three ways:

- Create a RejectedPromise, which only passes http exceptions to the callback, rethrows everything else
- When the RejectedPromise is created, make sure only HTTP exceptions are injected into the Promise (although others are possible too) and everything else is thrown immediately (this would require to modify adapter promise implementations too)
- Handle the scenario in the plugins: accept any exception and only process http ones, throw the rest (how about no?)

joelwurtz: > Handle the scenario in the plugins: accept any exception and only process http ones, throw the rest (how about no?)

No, or i do not maintain this xD, it will just be a nightmare to handle.

> Create a RejectedPromise, which only passes http exceptions to the callback, rethrows everything else
> When the RejectedPromise is created, make sure only HTTP exceptions are injected into the Promise (although others are possible too) and everything else is thrown immediately (this would require to modify adapter promise implementations too)

:+1: and we must do the same for the FulfilledPromise, as actually it's accept anything, where we should only allow `ResponseInterface`

joelwurtz: The PR is here https://github.com/php-http/httplug/pull/116

joelwurtz:
Should we close this as well since #42 has been merged ?

Nyholm:
Correct. Thank you

PR #38, opened 2016-08-12
[WIP] fix issue #37

### LINKED ISSUE #37: Redirect Plugin - Circular redirection detected
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 1.2.1

#### Actual Behavior
An `Http\Client\Common\Exception\CircularRedirectionException` is thrown when executing the following code:
```php
<?php

require 'vendor/autoload.php';

use Http\Adapter\Guzzle6\Client as HttpClient;
use Http\Client\Common\PluginClient;
use Http\Client\Common\Plugin\RedirectPlugin;
use Http\Message\MessageFactory\GuzzleMessageFactory as MessageFactory;

$message = new MessageFactory;

$pluginClient = new PluginClient(new HttpClient, [
    new RedirectPlugin,
]);

$request = $message->createRequest('GET', 'http://www.artstation.com/artwork.rss');
$response = $pluginClient->sendRequest($request);

var_dump($response->getBody()->getContents());
```

#### Expected Behavior
No exception should be thrown, since http://www.artstation.com/artwork.rss redirects to https://www.artstation.com/artwork.rss (`http` -> `https`)

#### Steps to Reproduce
Execute the code above.

#### Possible Solutions
The issue is on line [162](https://github.com/php-http/client-common/blob/master/src/Plugin/RedirectPlugin.php#L162), where the circular redirection detection is done.

This line of code:
```php
if (in_array($redirectRequest->getRequestTarget(), $this->circularDetection[$chainIdentifier]))
````
is checking if the current request target already exists in the circular detection map.

Because two lines before ([160](https://github.com/php-http/client-common/blob/master/src/Plugin/RedirectPlugin.php#L160)), the same value is being added, it will always be `true`.

### PR DESCRIPTION
**Pull request by quetzyg**
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #37
| Documentation   | N/A
| License         | MIT

#### What's in this PR?

The changes in this PR, allow the `RedirectPlugin` to properly track requests using full URLs instead of comparing only the URI.

#### Checklist

#### To Do

- [ ] Fix tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 ## 1.3.0 - Unreleased
 
+### Added
 - Add HttpClientPool client to leverage load balancing and fallback mechanism [see the documentation](http://docs.php-http.org/en/latest/components/client-common.html) for more details
+
+### Changed
+- RedirectPlugin: use the full URL instead of the URI to properly keep track of redirects
 
 ## 1.2.1 - 2016-07-26
 
--- a/src/Plugin/RedirectPlugin.php
+++ b/src/Plugin/RedirectPlugin.php
@@ -134,9 +134,9 @@ class RedirectPlugin implements Plugin
     public function handleRequest(RequestInterface $request, callable $next, callable $first)
     {
         // Check in storage
-        if (array_key_exists($request->getRequestTarget(), $this->redirectStorage)) {
-            $uri = $this->redirectStorage[$request->getRequestTarget()]['uri'];
-            $statusCode = $this->redirectStorage[$request->getRequestTarget()]['status'];
+        if (array_key_exists((string) $request->getUri(), $this->redirectStorage)) {
+            $uri = $this->redirectStorage[(string) $request->getUri()]['uri'];
+            $statusCode = $this->redirectStorage[(string) $request->getUri()]['status'];
             $redirectRequest = $this->buildRedirectRequest($request, $uri, $statusCode);
 
             return $first($redirectRequest);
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by dbu** [comment]
__toString is indeed not allowed to throw an exception - its unfortunate we don't see the stack trace.

is the test using a mock for UriInterface? could we use a real instance of Uri for the test instead?

**Comment by quetzyg** [comment]
I tried running with `-vvv`:
```sh
vendor/bin/phpspec -vvv spec/Plugin/RedirectPluginSpec.php
```
but there was no change in the output I got.

sagikazarmark:
I think the problem is that `__toString` is not expected to be called on the URI object. So far we used getRequestTarget:

https://github.com/php-http/client-common/blob/master/spec/Plugin/RedirectPluginSpec.php#L47

dbu:
mark, would you recommend changing the mock, or do you agree with me
that we better would use a real request implementation? there would be
no harm in adding `guzzlehttp/psr7` to require-dev...

sagikazarmark:
I would be happier if we could solve the promise/exception thing first.

sagikazarmark:
@quetzyg can you please check my recommended fix?

quetzyg:
Hi @sagikazarmark,

I'll try to find some time to make the changes to the spec expectations.

joelwurtz:
I've fixed test and add a specific one for the one mentioned in the issue under #43

dbu:
continued in #43. thanks @quetzyg for your contribution!

quetzyg:
Soz @dbu, I've been swamped and I haven't had time to check phpSpec properly to update the tests.
Glad @joelwurtz stepped up.

Keep up the good work guys!

PR #40, opened 2016-08-18
Added PluginClientBuilder

Nyholm:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Related tickets | related to https://github.com/php-http/HttplugBundle/issues/109 |
| Documentation |  |
| License | MIT |
#### What's in this PR?

This is a plugin builder that could be use to build up a Plugin client. See example usage: https://github.com/Nyholm/php-github-api/blob/plugin-client-builder/lib/Github/Client.php
#### Why?
- If a library builds up the PluginClient manually there is no way for us (HTTPlugBundle) to know what plugins that are used and we cannot show any plugin stack in the web profiler
- We could "risk" having multiple plugin chains which will cause issues when using the cache plugin if a later plugin is rewriting the url. (or any other part of the cache key)
- A pluginClientBuilder will make it easy for the LibraryBundle to integrate the library with Symfony and HTTPlugBundle. 
#### Checklist
#### To Do
- [ ] Clean the code
- [ ] Figure out if the interface is needed
- [ ] Figure out the final interface

This PR needs some work but I wanted to share this code

### DIFF
--- a/src/PluginClientBuilder.php
+++ b/src/PluginClientBuilder.php
@@ -1,0 +1,168 @@
+<?php
+
+namespace Http\Client\Common;
+
+use Http\Client\HttpClient;
+
+/**
+ * A builder that help you build a PluginClient.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class PluginClientBuilder implements PluginClientBuilderInterface
+{
+    /**
+     * @var Plugin[]
+     */
+    protected $plugins = [];
+
+    /**
+     * @var bool
+     */
+    protected $rebuildClient = true;
+
+    /**
+     * The last created client with the plugins
+     *
+     * @var PluginClient
+     */
+    private $pluginClient;
+
+    /**
+     * The object that sends HTTP messages
+     *
+     * @var HttpClient
+     */
+    private $httpClient;
+
+    /**
+     *
+     * @param HttpClient $httpClient
+     */
+    public function __construct(HttpClient $httpClient = null)
+    {
+        $this->httpClient = $httpClient;
+    }
+
+    /**
+     * Add a new plugin to the end of the plugin chain.
+     *
+     * @param Plugin $plugin
+     */
+    public function addPlugin(Plugin $plugin)
+    {
+        $this->plugins[] = $plugin;
+        $this->rebuildClient = true;
+    }
+
+    /**
+     * Remove a plugin by its fully qualified class name (FQCN).
+     *
+     * @param string $fqcn
+     */
+    public function removePlugin($fqcn)
+    {
+        foreach ($this->plugins as $idx => $plugin) {
+            if ($plugin instanceof $fqcn) {
+                unset($this->plugins[$idx]);
+                $this->rebuildClient = true;
+            }
+        }
+    }
+
+    /**
+     * Alias for removePlugin and addPlugin
+     *
+     * @param Plugin $plugin
+     */
+    public function replacePlugin(Plugin $plugin)
+    {
+        $this->removePlugin(get_class($plugin));
+        $this->addPlugin($plugin);
+    }
+
+    /**
+     * Get all plugins
+     *
+     * @return Plugin[]
+     */
+    public function getPlugins()
+    {
+        return $this->plugins;
+    }
+
+    /**
+     * This will overwrite all existing plugins
+     *
+     * @param Plugin[] $plugins
+     *
+     * @return PluginClientBuilder
+     */
+    public function setPlugins($plugins)
+    {
+        $this->plugins = $plugins;
+        $this->rebuildClient = true;
+    }
+
+    /**
+     * @return PluginClient
+     */
+    public function buildClient()
+    {
+        if (!$this->httpClient) {
+            throw new \RuntimeException('No HTTP client were provided to the PluginBuilder.');
+        }
+
+        if ($this->rebuildClient) {
+            $this->rebuildClient = false;
+            $this->pushBackCachePlugin();
+
+            $this->pluginClient = new PluginClient($this->httpClient, $this->plugins);
+        }
+
+        return $this->pluginClient;
+    }
+
+    /**
+     * @return boolean
+     */
+    public function isModified()
+    {
+        return $this->rebuildClient;
+    }
+
+    /**
+     * @param HttpClient $httpClient
+     */
+    public function setHttpClient(HttpClient $httpClient)
+    {
+        $this->rebuildClient = true;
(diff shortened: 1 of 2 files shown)


src/PluginClientBuilder.php
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /**
+     * Make sure to move the cache plugin to the end of the chain
+     */
+    private function pushBackCachePlugin()
+    {
+        $cachePlugin = null;
+        foreach ($this->plugins as $i => $plugin) {
+            if ($plugin instanceof Plugin\CachePlugin) {
```
GrahamCampbell: Might be worth adding an interface plugins can implement to say they need to get pushed back instead of hard coding the cache plugin like this?

src/PluginClientBuilder.php
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * Remove a plugin by its fully qualified class name (FQCN).
+     *
+     * @param string $fqcn
+     */
+    public function removePlugin($fqcn)
+    {
+        foreach ($this->plugins as $idx => $plugin) {
+            if ($plugin instanceof $fqcn) {
```
dbu: are you sure this works? i thought with strings, we need is_subclass_of?

src/PluginClientBuilder.php
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /**
+     * Alias for removePlugin and addPlugin
+     *
+     * @param Plugin $plugin
+     */
+    public function replacePlugin(Plugin $plugin)
+    {
+        $this->removePlugin(get_class($plugin));
```
dbu: what if there are several plugins with the same class ? we would remove all of them. not sure if that is a problem. i guess its the fuziness of builder.

we also don't respect the position, which might be more of an issue. could we put the plugin in the position where the one we remove was?

src/PluginClientBuilder.php
sagikazarmark: I am not sure I like the idea. Plugins should not be affected by external configuration logic. Also, there is no guarantee that cache should always be "pushed back".

src/PluginClientBuilder.php
```diff
@@ -0,0 +1,168 @@
...
+     *
+     * @var HttpClient
+     */
+    private $httpClient;
+
+    /**
+     *
+     * @param HttpClient $httpClient
+     */
+    public function __construct(HttpClient $httpClient = null)
```
dbu: why is that optional? this feels wrong if we don't use discovery for it.

src/PluginClientBuilder.php
sagikazarmark: I would rather use append/prependPlugin methods.

src/PluginClientBuilder.php
```diff
@@ -0,0 +1,168 @@
...
+        $this->plugins = $plugins;
+        $this->rebuildClient = true;
+    }
+
+    /**
+     * @return PluginClient
+     */
+    public function buildClient()
+    {
+        if (!$this->httpClient) {
```
dbu: we should make sure in the constructor that this can not happen

src/PluginClientBuilder.php
sagikazarmark: I think it does. You need is_subclass_of when you don't have an instance, just the class name.

src/PluginClientBuilder.php
```diff
@@ -0,0 +1,168 @@
...
+    /**
+     * Make sure to move the cache plugin to the end of the chain
+     */
+    private function pushBackCachePlugin()
+    {
+        $cachePlugin = null;
+        foreach ($this->plugins as $i => $plugin) {
+            if ($plugin instanceof Plugin\CachePlugin) {
+                $cachePlugin = $plugin;
+                unset($this->plugins[$i]);
```
dbu: this could break in some edge cases when people have more than one cache plugin, maybe extending the base class to cache some requests differently. 

if somebody would do the same with RequestMatcherPlugin, the cache would not be moved to the end at all.

Nyholm:
I agree with all your comments. I created this by extracting logic from the KnpGithub client. I'll adapt to your feedback to make this generic.

src/PluginClientBuilder.php
dbu: maybe we should not have this method but have people get the plugins, change the list and set plugins. RequestMatcherPlugin for example is quite likely to be used multiple times.

src/PluginClientBuilder.php
dbu: if we find a way to make replace preserve the order, we would not need this.

src/PluginClientBuilder.php
GrahamCampbell: I agree. I'd say make adding irrevocable.

dbu:
okay, the knp example is a very good illustration where this builder approach works very well.

some braindump trying to figure out what we should do:

in a symfony context, a compiler pass to replace services with other services would be much cleaner, and would allow to leverage the configuration options on httplug bundle. but that would mean you need a way to set up the library without framework, and a different set up for each framework integration.

the thing i don't like on the knp library is that Client is a mix of an api client and a http client builder. your refactoring pushes it even more in that direction. with symfony, we would be able to change the list of plugins in a compiler pass at container build time (aka only on cache rebuild). with the builder in php, instances are created, some potentially only to be discarded later.

is it the idea of that library that you would change the http setup while communicating with github? as dev, i would prefer several client instances if i need different http behaviour for different tasks, like using different credentials. its way to easy otherwise to mix things up and use the wrong settings.

my idea would be that for a frameworkagnostic library, you have a builder (or whatever) that gives you the github client instance, shuffling around plugins. but for symfony, it would make more sense to not use that and have a compiler pass to manipulate the plugins of the client service that is specified in the configuration. this could get hairy when several bundles use the same client, for example the default client, but would be more efficient to instantiate the client, and easier for the user to configure.

dbu:
while preparing the web summer camp tutorial, i had a completely different idea: what if the HttplugBundle would offer a method that accepts a client configuration as the bundle config would do it, and returns a service definition? then a bundle could do its own configuration with extra plugins and whatnot and then add the plugin configuration/services it needs, do custom validation or whatever and still get a properly set up service.

this approach would be specific to symfony - as are bundles in general. for other frameworks, we would need to find an approach suitable for those frameworks. but maybe its better to have framework integration framework specific?

fbourigault:
Since we have #71 and a plan for https://github.com/php-http/HttplugBundle/issues/109, I think we could close this PR.

PR #42, opened 2016-08-31
Use http based promise in emulated trait and specs

### LINKED ISSUE #34: Clear exception usage in Plugins

joelwurtz:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #34 , replace #35
| Documentation   | /
| License         | MIT

See issue #34 and discussion in #35 also for more information

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     ],
     "require": {
         "php": ">=5.4",
-        "php-http/httplug": "^1.0",
+        "php-http/httplug": "^1.1",
         "php-http/message-factory": "^1.0",
         "php-http/message": "^1.2",
         "symfony/options-resolver": "^2.6 || ^3.0"
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -20,6 +20,8 @@ interface Plugin
     /**
      * Handle the request and return the response coming from the next callable.
      *
+     * @see http://docs.php-http.org/en/latest/plugins/build-your-own.html
+     *
      * @param RequestInterface $request
      * @param callable         $next    Next middleware in the chain, the request is passed as the first argument
      * @param callable         $first   First middleware in the chain, used to to restart a request
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -24,7 +24,7 @@ interface Plugin
      * @param callable         $next    Next middleware in the chain, the request is passed as the first argument
      * @param callable         $first   First middleware in the chain, used to to restart a request
      *
-     * @return Promise
+     * @return Promise Resolves a PSR-7 Response or fails with an Http\Client\Exception (The same as HttpAsyncClient).
      */
     public function handleRequest(RequestInterface $request, callable $next, callable $first);
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Add HttpClientPool client to leverage load balancing and fallback mechanism [see the documentation](http://docs.php-http.org/en/latest/components/client-common.html) for more details
 
 ### Changed
+- Fix Emulated Trait to use Http based promise which respect the HttpAsyncClient interface
 - RedirectPlugin: use the full URL instead of the URI to properly keep track of redirects
 
 ## 1.2.1 - 2016-07-26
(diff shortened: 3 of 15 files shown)


composer.json
```diff
@@ -12,7 +12,7 @@
     ],
     "require": {
         "php": ">=5.4",
-        "php-http/httplug": "^1.0",
+        "php-http/httplug": "^1.1.0",
```
GrahamCampbell: `^1.1` would be fine here

joelwurtz: Yes, it's better, thanks.

GrahamCampbell:
Ok, so what happens when a non-http exception gets thrown?

Nyholm:
When a non-httplug exception is thrown we abort the plugin chain and the exception bubble up like normal. Only httplug exception will respect the plugin chain.

GrahamCampbell:
This will break other libraries though.

sagikazarmark:
I don't think so. If you used a plugin with rejection callbacks broke your library too. This was a bug IMO, which has been solved. Also, this bug only affects you if you use plugins.

src/Plugin.php
```diff
@@ -24,7 +24,7 @@
      * @param callable         $next    Next middleware in the chain, the request is passed as the first argument
      * @param callable         $first   First middleware in the chain, used to to restart a request
      *
-     * @return Promise
+     * @return Promise Resolves a PSR-7 Response or fails with an Http\Client\Exception (The same as HttpAsyncClient).
```
joelwurtz: Is this a BC break ? (For me no, it's more a clarification but i dont know if this interface is already used but other packages than one we have under control ?)

sagikazarmark:
I added some explanation here: https://github.com/php-http/httplug/releases/tag/v1.1.0

src/Plugin.php
Nyholm: Could we improve this doc block somehow even more?

Maybe describe what the callables will return?

joelwurtz: Don't we have already this on the documentation ? if so would just add a link with `@see`

Nyholm: We don't but we should. Write 
`@see http://php-http.readthedocs.io/en/latest/plugins/introduction.html`

FYI: https://github.com/php-http/documentation/issues/149

Nyholm:
Thank you @joelwurtz. 

Is this ready for merge?

src/Plugin.php
sagikazarmark: From my POV, it's not. Http Promises still fullfill the Promise interface.

sagikazarmark:
Please gimme a note when this can be merged and tagged because I would like to do some other things too.

src/Plugin.php
Nyholm: Add "@see http://docs.php-http.org/en/latest/plugins/build-your-own.html". I've added some lines about exceptions (https://github.com/php-http/documentation/pull/151)

> Is this a BC break ?

No, this is not a BC break.

joelwurtz:
For me all the work is done here, so feel free to merge if it's ok to you (not a fan of merging my own PR in a collaboration project, it feels presumptuous...)

Nyholm:
Thank you for this PR. Good work!

Issue #41, opened 2016-08-23
Tag 1.3.0?


Nyholm:
Are we happy with the ClientPool? Are we confident to tag 1.3.0?

joelwurtz:
Is it possible to add a experimental flag on ClientPool ? In fact i'm not still confident on the public api exposed and the way we can extends the Pool i would prefer to have something that can be broken on HttpClientPool but not quite sure on how to reflect that ?

sagikazarmark:
I am happy with the current state.

@joelwurtz Do you think ClientPool is done and ready for release?

sagikazarmark:
I see. Well, I would be happier if we could avoid releasing experimental stuff, but the HTTP Promise thing should be released as soon as possible. Is there a list of issues that we should still solve/discuss for the client pool?

Based on that we could decide whether we want to wait for the Pool clients to get more mature or just mark them as experimental and release the next version.

joelwurtz:
or just mark all methods inside as `@internal` ?

sagikazarmark:
That's okay from a logical point of view, but could still lead to code breaks which I would like to avoid. That said, we can make that call, but I would like to see the current state clearly.

Nyholm:
> or just mark all methods inside as @internal ?

And the class. 

One could also release a version without the ClientPool.

GrahamCampbell:
I'd say tag this thing. 👍

acrobat:
Any news on an eta for the 1.3 release?

sagikazarmark:
As far as I can see this is just about to happen.

Nyholm:
Yes. We will tag 1.3.0 this weekend.

PR #47, opened 2016-10-13
Removed the HTTPClientPool feature

Nyholm:
This PR removes the HttpClientPool feature from master which would allow us to tag 1.3.0 with confidence. After this PR is merged I will create a feature branch that reverts commit 51273fc and we can continue to develop and test the client pool until we are confident to include it in a release. 

@joelwurtz, I would of course like to hear your thoughts about this solution.  

Related to #41


dbu:
lets do this so we can move forward.

Nyholm:
Thank you for merging

PR #48, opened 2016-10-14
Add AddPathPlugin and BaseUriPlugin

Soullivaneuh:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no since this is a kind of bug regarding the situation
| Deprecations?   | no
| Related tickets | N/A
| Documentation   | https://github.com/php-http/documentation/pull/163
| License         | MIT

#### What's in this PR?

This bugfix permits to keep the base path of an URI on the `AddHost` plugin.

#### Why?

With the following PHP code example:

```php
$httpClient = new HttpMethodsClient(
    new PluginClient(HttpClientDiscovery::find(), [
        new AddHostPlugin(UriFactoryDiscovery::find()->createUri('https://logs.domain.com/api')),
        new ErrorPlugin(),
    ]),
    MessageFactoryDiscovery::find()
);
$httpClient->get('/users');
```

The https://logs.domain.com/users URL will be called instead of /api/users.

The base path should be kept on this plugin.

#### Checklist

- [x] Update the tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Require Httplug 1.1 where we use HTTP specific promises. 
 - RedirectPlugin: use the full URL instead of the URI to properly keep track of redirects
 - Add AddPathPlugin for API URLs with base path
+- Add BaseUriPlugin that combines AddHostPlugin and AddPathPlugin
 
 ## 1.2.1 - 2016-07-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Fix Emulated Trait to use Http based promise which respect the HttpAsyncClient interface
 - Require Httplug 1.1 where we use HTTP specific promises. 
 - RedirectPlugin: use the full URL instead of the URI to properly keep track of redirects
-
+- Add AddPathPlugin for API URLs with base path
 
 ## 1.2.1 - 2016-07-26
 
--- a/spec/Plugin/AddPathPluginSpec.php
+++ b/spec/Plugin/AddPathPluginSpec.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace spec\Http\Client\Common\Plugin;
+
+use Http\Message\StreamFactory;
+use Http\Message\UriFactory;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\UriInterface;
+use PhpSpec\ObjectBehavior;
+
+class AddPathPluginSpec extends ObjectBehavior
+{
+    function let(UriInterface $uri)
+    {
+        $this->beConstructedWith($uri);
+    }
+
+    function it_is_initializable(UriInterface $uri)
+    {
+        $uri->getPath()->shouldBeCalled()->willReturn('/api');
+
+        $this->shouldHaveType('Http\Client\Common\Plugin\AddPathPlugin');
+    }
+
+    function it_is_a_plugin(UriInterface $uri)
+    {
+        $uri->getPath()->shouldBeCalled()->willReturn('/api');
+
+        $this->shouldImplement('Http\Client\Common\Plugin');
+    }
+
+    function it_adds_path(
+        RequestInterface $request,
+        UriInterface $host,
+        UriInterface $uri
+    ) {
+        $host->getPath()->shouldBeCalled()->willReturn('/api');
+
+        $request->getUri()->shouldBeCalled()->willReturn($uri);
+        $request->withUri($uri)->shouldBeCalled()->willReturn($request);
+
+        $uri->withPath('/api/users')->shouldBeCalled()->willReturn($uri);
+        $uri->getPath()->shouldBeCalled()->willReturn('/users');
+
+        $this->beConstructedWith($host);
+        $this->handleRequest($request, function () {}, function () {});
+    }
+
+    function it_throws_exception_on_trailing_slash(UriInterface $host)
+    {
+        $host->getPath()->shouldBeCalled()->willReturn('/api/');
+
+        $this->beConstructedWith($host);
+        $this->shouldThrow('\LogicException')->duringInstantiation();
+    }
+
+    function it_throws_exception_on_empty_path(UriInterface $host)
+    {
+        $host->getPath()->shouldBeCalled()->willReturn('');
+
+        $this->beConstructedWith($host);
+        $this->shouldThrow('\LogicException')->duringInstantiation();
+    }
+}
--- a/spec/Plugin/BaseUriPluginSpec.php
+++ b/spec/Plugin/BaseUriPluginSpec.php
@@ -1,0 +1,102 @@
+<?php
+
+namespace spec\Http\Client\Common\Plugin;
+
+use Http\Message\StreamFactory;
+use Http\Message\UriFactory;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\UriInterface;
+use PhpSpec\ObjectBehavior;
+
+class BaseUriPluginSpec extends ObjectBehavior
+{
+    function let(UriInterface $uri)
+    {
+        $this->beConstructedWith($uri);
+    }
+
+    function it_is_initializable(UriInterface $uri)
+    {
+        $uri->getHost()->shouldBeCalled()->willReturn('example.com');
+        $uri->getPath()->shouldBeCalled()->willReturn('/api');
+
+        $this->shouldHaveType('Http\Client\Common\Plugin\BaseUriPlugin');
+    }
+
+    function it_is_a_plugin(UriInterface $uri)
+    {
+        $uri->getHost()->shouldBeCalled()->willReturn('example.com');
+        $uri->getPath()->shouldBeCalled()->willReturn('/api');
+
+        $this->shouldImplement('Http\Client\Common\Plugin');
+    }
+
+    function it_adds_domain_and_path(
+        RequestInterface $request,
+        UriInterface $host,
+        UriInterface $uri
+    ) {
+        $host->getScheme()->shouldBeCalled()->willReturn('http://');
+        $host->getHost()->shouldBeCalled()->willReturn('example.com');
+        $host->getPort()->shouldBeCalled()->willReturn(8000);
+        $host->getPath()->shouldBeCalled()->willReturn('/api');
+
+        $request->getUri()->shouldBeCalled()->willReturn($uri);
+        $request->withUri($uri)->shouldBeCalled()->willReturn($request);
+
+        $uri->withScheme('http://')->shouldBeCalled()->willReturn($uri);
+        $uri->withHost('example.com')->shouldBeCalled()->willReturn($uri);
+        $uri->withPort(8000)->shouldBeCalled()->willReturn($uri);
+        $uri->withPath('/api/users')->shouldBeCalled()->willReturn($uri);
+        $uri->getHost()->shouldBeCalled()->willReturn('');
+        $uri->getPath()->shouldBeCalled()->willReturn('/users');
+
+        $this->beConstructedWith($host);
+        $this->handleRequest($request, function () {}, function () {});
+    }
+
+    function it_adds_domain(
+        RequestInterface $request,
+        UriInterface $host,
+        UriInterface $uri
+    ) {
+        $host->getScheme()->shouldBeCalled()->willReturn('http://');
+        $host->getHost()->shouldBeCalled()->willReturn('example.com');
+        $host->getPort()->shouldBeCalled()->willReturn(8000);
+        $host->getPath()->shouldBeCalled()->willReturn('/');
+
+        $request->getUri()->shouldBeCalled()->willReturn($uri);
+        $request->withUri($uri)->shouldBeCalled()->willReturn($request);
+
+        $uri->withScheme('http://')->shouldBeCalled()->willReturn($uri);
+        $uri->withHost('example.com')->shouldBeCalled()->willReturn($uri);
+        $uri->withPort(8000)->shouldBeCalled()->willReturn($uri);
+        $uri->getHost()->shouldBeCalled()->willReturn('');
+
+        $this->beConstructedWith($host);
+        $this->handleRequest($request, function () {}, function () {});
+    }
+
+    function it_replaces_domain_and_adds_path(
+        RequestInterface $request,
+        UriInterface $host,
+        UriInterface $uri
+    ) {
+        $host->getScheme()->shouldBeCalled()->willReturn('http://');
+        $host->getHost()->shouldBeCalled()->willReturn('example.com');
+        $host->getPort()->shouldBeCalled()->willReturn(8000);
+        $host->getPath()->shouldBeCalled()->willReturn('/api');
+
+        $request->getUri()->shouldBeCalled()->willReturn($uri);
+        $request->withUri($uri)->shouldBeCalled()->willReturn($request);
+
+        $uri->withScheme('http://')->shouldBeCalled()->willReturn($uri);
+        $uri->withHost('example.com')->shouldBeCalled()->willReturn($uri);
+        $uri->withPort(8000)->shouldBeCalled()->willReturn($uri);
+        $uri->withPath('/api/users')->shouldBeCalled()->willReturn($uri);
+        $uri->getPath()->shouldBeCalled()->willReturn('/users');
+
+        $this->beConstructedWith($host, ['replace' => true]);
+        $this->handleRequest($request, function () {}, function () {});
+    }
+}
--- a/src/Plugin/AddPathPlugin.php
+++ b/src/Plugin/AddPathPlugin.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Http\Client\Common\Plugin;
+
+use Http\Client\Common\Plugin;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\UriInterface;
+
+/**
+ * Prepend a base path to the request URI. Useful for base API URLs like http://domain.com/api.
+ *
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+final class AddPathPlugin implements Plugin
+{
+    /**
+     * @var UriInterface
+     */
+    private $uri;
+
+    /**
+     * @param UriInterface $uri
+     */
+    public function __construct(UriInterface $uri)
+    {
+        if ($uri->getPath() === '') {
+            throw new \LogicException('URI path cannot be empty');
+        }
+
+        if (substr($uri->getPath(), -1) === '/') {
+            throw new \LogicException('URI path cannot end with a slash.');
+        }
+
+        $this->uri = $uri;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
+        $request = $request->withUri($request->getUri()
+            ->withPath($this->uri->getPath().$request->getUri()->getPath())
+        );
+
+        return $next($request);
+    }
+}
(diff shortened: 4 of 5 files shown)


dbu:
as far as i remember, @sagikazarmark had a good argument why this should be a separate plugin. i can't find the place we discussed this anymore. mark would you rembember?

doing this like you do here would be a BC break-ish thing as people that now use a domain with path and rely on the path not being appended would see their applications stop working. and its a new feature, not an actual bugfix so i think we should not do a BC break in a minor version for this.

for the content, question is whether we treat `/path/something` and `path/something` the same or only prepend a path when there is no leading `/`? depending on the scenario, both can make sense.

sagikazarmark:
This one? https://github.com/php-http/plugins/pull/42#issuecomment-171611857

dbu:
exactly, thanks!

to sum it up:

separate plugin for path: 
* + clearly separated logic
* + mix and match with replace host or add host
* - two plugins to configure
* - unexpected behaviour when specifying path to AddHostPlugin => should we forbid that?

if we would mix it together, would we only add the path if the host is missing, or always even if the host exists? should we check if the path already starts with the configured path and if so, skip adding it again (probably not)?

@Soullivaneuh what is your opinion on the other thread? would you agree that we move this to a separate plugin?

sagikazarmark:
AddHostPlugin only sets the host, scheme and port, ommits the path at the moment.

joelwurtz:
Yes we talk about this and say it must be in a separate plugin like an `AddPathPlugin`, then we could imagine a third one using these two (like `BaseUriPlugin`).

Soullivaneuh:
> doing this like you do here would be a BC break-ish thing as people that now use a domain with path and rely on the path not being appended would see their applications stop working. and its a new feature, not an actual bugfix so i think we should not do a BC break in a minor version for this.

@dbu This is *technicaly* a BC break, and you know how I'm pernickety with this for Sonata. :wink:
But for me it's not really a BC break because who will setup `http://domain.com/api` on this plugin if the `/api` append does not work?

But I agree with your point of you about the goal of the plugin, I hesitated for this too.

I'll rework it for a `AddPathPlugin` then.

> then we could imagine a third one using these two (like BaseUriPlugin).

@joelwurtz What do you mean by "using"? I don't see how expect rewrite the logic of both on this class.

joelwurtz:
> @joelwurtz What do you mean by "using"? I don't see how expect rewrite the logic of both on this class.

Something like this : 

```php

class BaseUriPlugin
{  
    private $addHostPlugin;

    private $addPathPlugin;

    public function __construct($options = [])
    {
        // @TODO deal with options
        $this->addHostPlugin = new AddHostPlugin($options);
        $this->addPathPlugin = new AddPathPlugin($options);
    }

    {
        $addHostNext = function (Request $request) use ($next, $first) {
             return $this->addHostPlugin->handleRequest($request, $next, $first);
        };

        return $this->addPathPlugin->handleRequest($request, $addHostNext, $first);
    }
}
```

dbu:
i like joel's proposal. the coupling of BaseUriPlugin will be a bit tight like this. Maybe use two option arguments, one for host and one for path? that leaks the abstraction but is clean and future compatible

Soullivaneuh:
> Maybe use two option arguments, one for host and one for path?

My `AddPathPlugin` does not need options ATM.

Soullivaneuh:
Code updated, thanks for reviewing.

src/Plugin/BaseUriPlugin.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        $addHostNext = function (RequestInterface $request) use ($next, $first) {
+            return $this->addHostPlugin->handleRequest($request, $next, $first);
+        };
+
+        if ($this->addPathPlugin) {
+            return $this->addPathPlugin->handleRequest($request, $addHostNext, $first);
+        }
+
+        return $addHostNext;
```
joelwurtz: Should be `return $addHostNext($request);`

spec/Plugin/BaseUriPluginSpec.php
```diff
@@ -0,0 +1,80 @@
...
+
+        $uri->withScheme('http://')->shouldBeCalled()->willReturn($uri);
+        $uri->withHost('example.com')->shouldBeCalled()->willReturn($uri);
+        $uri->withPort(8000)->shouldBeCalled()->willReturn($uri);
+        $uri->withPath('/api/users')->shouldBeCalled()->willReturn($uri);
+        $uri->getPath()->shouldBeCalled()->willReturn('/users');
+
+        $this->beConstructedWith($host, ['replace' => true]);
+        $this->handleRequest($request, function () {}, function () {});
+    }
```
joelwurtz: Can you add test when there is no path ? (with my previous comment it should have failed :) )

src/Plugin/BaseUriPlugin.php
Soullivaneuh: Fixed.

spec/Plugin/BaseUriPluginSpec.php
Soullivaneuh: Fixed.

src/Plugin/AddPathPlugin.php
```diff
@@ -0,0 +1,44 @@
...
+     */
+    private $host;
+
+    /**
+     * @param UriInterface $host
+     */
+    public function __construct(UriInterface $host)
+    {
+        if ($host->getPath() === '') {
+            throw new \LogicException('Host path can not be empty');
```
xabbuh: cannot

src/Plugin/AddPathPlugin.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    /**
+     * @var UriInterface
+     */
+    private $host;
+
+    /**
+     * @param UriInterface $host
+     */
+    public function __construct(UriInterface $host)
```
xabbuh: By the way, what is the advantage of passing an `UriInterface` instance here instead of a plain string?

Nyholm: Do we really? the only with we do with $host is to convert it to a string..

src/Plugin/BaseUriPlugin.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+final class BaseUriPlugin implements Plugin
+{
+    /**
+     * @var AddHostPlugin
+     */
+    private $addHostPlugin;
+
+    /**
+     * @var AddPathPlugin
```
dbu: this is AddPathPlugin|null

src/Plugin/AddPathPlugin.php
xabbuh: Well, actually we don't need a host here at all. The user just has to pass a path (not a host) and I am not sure that actually needs any validation at all.

dbu: no, we do `$host->getPath()` once here to validate, then again where we use it to just get the path part. if we would accept any string, you could create this plugin with "foo bar baz" and we would try to change the request uri to that. or we start to parse urls in here (which might have been validated and parsed before being passed), which is what the Uri model would free us from.

src/Plugin/BaseUriPlugin.php
xabbuh: I do not think the docblock here is necessary at all. IDEs should be able to infer the type themselves (the same applies to the  `$addHostPlugin` property).

src/Plugin/BaseUriPlugin.php
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * @param UriInterface $host
+     * @param array        $hostConfig Config for AddHostPlugin. @see AddHostPlugin::configureOptions
+     */
+    public function __construct(UriInterface $host, array $hostConfig = [])
+    {
+        $this->addHostPlugin = new AddHostPlugin($host, $hostConfig);
+
+        if ($host->getPath()) {
```
dbu: we should rtrim($path, '/') here. or throw an exception on http://domain.com/ (it makes no sense to prepend just / to a uri)

src/Plugin/AddPathPlugin.php
Soullivaneuh: I agree with @dbu.

This also make an easier argument passing for `BaseUriPlugin`.

After that, it's as you want. I can pass a string instead if you prefer.

Soullivaneuh: Discussed here: https://github.com/php-http/client-common/pull/48#discussion_r83790722

src/Plugin/AddPathPlugin.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->host = $host;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
+        $request = $request->withUri($request->getUri()
+            ->withPath($this->host->getPath().$request->getUri()->getPath())
```
dbu: what about anything ending on /? notably http://domain.com/ but also http://domain.com/path/ - both would end up in double leading slash, which is very likely not what the user wanted.

this is a rather annoying edge case. i think we should somehow handle it (maybe validate and InvalidArgumentException) to avoid wtf moments (some servers will handle the example.com//something paths, others wont). if somebody actually wants to create double slashes, they can write their own plugin, its a rare edgecase.

Soullivaneuh: I will add a test for that and handle it.

src/Plugin/AddPathPlugin.php
Soullivaneuh: Fixed.

src/Plugin/BaseUriPlugin.php
Soullivaneuh: Fixed.

src/Plugin/BaseUriPlugin.php
Soullivaneuh: Fixed.

src/Plugin/AddPathPlugin.php
Soullivaneuh: Fixed.

Soullivaneuh:
Requested changes was made.

Except if you want me to change something on https://github.com/php-http/client-common/pull/48/files/2aac11b11bfd025c9d6082e722a1408139ab36f8#diff-44d3f10fc0543e373abc05223f369e01, this PR should be LGTM.

CHANGELOG.md
```diff
@@ -8,7 +8,8 @@
 - Fix Emulated Trait to use Http based promise which respect the HttpAsyncClient interface
 - Require Httplug 1.1 where we use HTTP specific promises. 
 - RedirectPlugin: use the full URL instead of the URI to properly keep track of redirects
-
+- Add AddPathPlugin for API URLs with base path
+- Combine AddHostPlugin and AddPathPlugin onto BaseUriPlugin
```
dbu: Lets say "Add BaseUriPlugin that combines ..."

src/Plugin/AddPathPlugin.php
```diff
@@ -0,0 +1,44 @@
...
+     * @var UriInterface
+     */
+    private $host;
+
+    /**
+     * @param UriInterface $host
+     */
+    public function __construct(UriInterface $host)
+    {
+        if (rtrim($host->getPath(), '/') === '') {
```
dbu: i think i would check '' or last character is '/' and throw an exception when it is. silently fixing things makes harder debugging. after doing example.com/ once or twice, people will learn to use example.com

src/Plugin/AddPathPlugin.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->host = $host;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
+        $request = $request->withUri($request->getUri()
+            ->withPath(str_replace('//', '/', $this->host->getPath().$request->getUri()->getPath()))
```
dbu: this feels odd. i would rtrim '/' again. if somebody puts double slashes elsewhere, they probably wanted that. but see above...

src/Plugin/AddPathPlugin.php
```diff
@@ -0,0 +1,44 @@
...
+ * Add base path to the request URI. Useful for base API URLs like http://domain.com/api.
+ *
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+final class AddPathPlugin implements Plugin
+{
+    /**
+     * @var UriInterface
+     */
+    private $host;
```
dbu: lets call this $uri, instead of $host. also in the constructor argument.

src/Plugin/BaseUriPlugin.php
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * @var AddPathPlugin|null
+     */
+    private $addPathPlugin = null;
+
+    /**
+     * @param UriInterface $host
+     * @param array        $hostConfig Config for AddHostPlugin. @see AddHostPlugin::configureOptions
+     */
+    public function __construct(UriInterface $host, array $hostConfig = [])
```
dbu: s/$host/$uri/

src/Plugin/BaseUriPlugin.php
```diff
@@ -0,0 +1,54 @@
...
+     */
+    private $addHostPlugin;
+
+    /**
+     * @var AddPathPlugin|null
+     */
+    private $addPathPlugin = null;
+
+    /**
+     * @param UriInterface $host
```
dbu: we should say that this must contain a host name and may contain a path that must not end on /

src/Plugin/AddPathPlugin.php
Soullivaneuh: If we throw an exception like asked in https://github.com/php-http/client-common/pull/48#discussion_r83802537, we don't need to rtrim anything IMHO.

Soullivaneuh:
@dbu All requested changes was made.

src/Plugin/AddPathPlugin.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace Http\Client\Common\Plugin;
+
+use Http\Client\Common\Plugin;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\UriInterface;
+
+/**
+ * Add base path to the request URI. Useful for base API URLs like http://domain.com/api.
```
Nyholm: Maybe be more clear here: 

"Prepend a base path to..."

dbu:
LGTM

src/Plugin/BaseUriPlugin.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+namespace Http\Client\Common\Plugin;
+
+use Http\Client\Common\Plugin;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\UriInterface;
+
+/**
+ * This plugin combines the host and path plugins.
```
Nyholm: Maybe: 
This plugin combines the AddHostPlugin and AddPathPlugin.

spec/Plugin/AddPathPluginSpec.php
```diff
@@ -0,0 +1,66 @@
...
+class AddPathPluginSpec extends ObjectBehavior
+{
+    function let(UriInterface $uri)
+    {
+        $this->beConstructedWith($uri);
+    }
+
+    function it_is_initializable(UriInterface $uri)
+    {
+        $uri->getPath()->shouldBeCalled()->willReturn('/api');
```
sagikazarmark: I think this could be moved to the let method. That would simplify things.

spec/Plugin/AddPathPluginSpec.php
```diff
@@ -0,0 +1,66 @@
...
+        UriInterface $host
+    ) {
+        $host->getPath()->shouldBeCalled()->willReturn('/api/');
+
+        $this->beConstructedWith($host);
+        $this->shouldThrow('\LogicException')->duringInstantiation();
+    }
+
+    function it_throws_exception_on_empty_path(
+        UriInterface $host
```
sagikazarmark: Why multiline?

spec/Plugin/AddPathPluginSpec.php
```diff
@@ -0,0 +1,66 @@
...
+
+        $uri->withPath('/api/users')->shouldBeCalled()->willReturn($uri);
+        $uri->getPath()->shouldBeCalled()->willReturn('/users');
+
+        $this->beConstructedWith($host);
+        $this->handleRequest($request, function () {}, function () {});
+    }
+
+    function it_throws_exception_on_trailing_slash(
+        UriInterface $host
```
sagikazarmark: Why multiline?

spec/Plugin/BaseUriPluginSpec.php
```diff
@@ -0,0 +1,102 @@
...
+class BaseUriPluginSpec extends ObjectBehavior
+{
+    function let(UriInterface $uri)
+    {
+        $this->beConstructedWith($uri);
+    }
+
+    function it_is_initializable(UriInterface $uri)
+    {
+        $uri->getHost()->shouldBeCalled()->willReturn('example.com');
```
sagikazarmark: Same here: can these be moved to let?

Soullivaneuh: I tried this:

```php
    function let(UriInterface $uri)
    {
        $uri->getPath()->shouldBeCalled()->willReturn('/api');

        $this->beConstructedWith($uri);
    }
```

But got this error:

```

        Http/Client/Common/Plugin/AddPathPlugin
  47  ✘ throws exception on trailing slash
        some predictions failed:
          Double\UriInterface\P114:
            No calls have been made that match:
              Double\UriInterface\P114->getPath()
            but expected at least one.

        Http/Client/Common/Plugin/AddPathPlugin
  56  ✘ throws exception on empty path
        some predictions failed:
          Double\UriInterface\P116:
            No calls have been made that match:
              Double\UriInterface\P116->getPath()
            but expected at least one.
```

I don't think this can be refactored on the `let` method because the return value changes on some cases.

Soullivaneuh:
@Nyholm @sagikazarmark Requested changes made, except https://github.com/php-http/client-common/pull/48#discussion_r83830214 for the moment.

dbu:
great! there is one thing left to do: the documentation pull request so that these plugins are documented as well. can you please do that?

Soullivaneuh:
Sure! Shall I just add two files on this part? https://github.com/php-http/documentation/tree/master/plugins

Nyholm:
Yes, I've added an issue for that. See https://github.com/php-http/documentation/issues/162

spec/Plugin/BaseUriPluginSpec.php
joelwurtz: AFAIK, let will set default config, so you can set the default input in the  ̀let` method and override it where it needs (when there is an empty path by example)

Soullivaneuh:
Documentation added: https://github.com/php-http/documentation/pull/163

dbu:
thanks @Soullivaneuh !

sagikazarmark:
Thank you @Soullivaneuh

Soullivaneuh:
Thank you too! Could we please have a new release for this feature? Thanks.

sagikazarmark:
Done: https://github.com/php-http/client-common/releases/tag/v1.3.1

xabbuh:
Shouldn't that have become 1.4.0?

And by the way, could you fix the link to the changelog for the 1.3.0 release? :)

Soullivaneuh:
> Shouldn't that have become 1.4.0?

I agree, it's a new feature. It should be a new minor if you follow semver ofc.

sagikazarmark:
Well, it didn't seem to be that huge change, but you are right.

sagikazarmark:
Done

Soullivaneuh:
Thank you!

PR #50, opened 2016-10-18
Php cs fixer bridge requirement

Soullivaneuh:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| Related tickets | https://github.com/php-http/client-common/commit/87c3bcd5c11c5c9f956ef833e25641a9d8ce4094#commitcomment-19467393 |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /phpspec.yml
 /phpunit.xml
 /vendor/
+/.php_cs.cache
--- a/.php_cs
+++ b/.php_cs
@@ -1,12 +1,13 @@
 <?php
 
 /*
- * In order to make it work, fabpot/php-cs-fixer and sllh/php-cs-fixer-styleci-bridge must be installed globally
- * with composer.
+ * In order to make it work, fabpot/php-cs-fixer must be installed globally with composer.
  *
  * @link https://github.com/Soullivaneuh/php-cs-fixer-styleci-bridge
  * @link https://github.com/FriendsOfPHP/PHP-CS-Fixer
  */
+
+require_once __DIR__.'/vendor/sllh/php-cs-fixer-styleci-bridge/autoload.php';
 
 use SLLH\StyleCIBridge\ConfigBridge;
 
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
     },
     "require-dev": {
         "phpspec/phpspec": "^2.4",
-        "henrikbjorn/phpspec-code-coverage" : "^1.0"
+        "henrikbjorn/phpspec-code-coverage" : "^1.0",
+        "sllh/php-cs-fixer-styleci-bridge": "^2.1"
     },
     "suggest": {
         "php-http/logger-plugin": "PSR-3 Logger plugin",


sagikazarmark:
Thanks @Soullivaneuh.

@joelwurtz WDYT?

joelwurtz:
> Yes. I'm against including php-cs-fixer too. I'm talking about the bridge library, which is... a library.

Which is a library for the need of extending php-cs-fixer so dev can use the style ci file but it's not in any way linked to php-http/client-common.

Our dependency is only on the contribution guide saying that we use PSR2 coding standard, the style ci file is only for our continuous integration pipeline and auto checking of the PR. 

Having php-cs-fixer is something that some of us use to auto correct the code we write (i'm using it as a trigger on my IDE) and the bridge is only a suggestion for people wanting to use php-cs-fixer and be compliant with the defined standard without having to read the styleci file or our contributing guide.

Adding this library is our dev dependencies add a maintenance cost for a suggestion of a suggestion... 

I'm more in favor of removing the documentation and let the user do what it wants to support psr2 than adding this.

sagikazarmark:
Alright, as usual, I am putting this to a vote: @php-http/owners @php-http/httplug

Nyholm:
Im not too happy with a code checking tool as a dependency. Nothing **in** our code relates to this tool. This tool is not a dependency but something developers/CI uses to check the quality of the code.

👎 

(I like that you raised the question because the answer is not obvious.)

sagikazarmark:
@Nyholm @jeromegamez can you please vote with thums up/down on my comment too?

jeromegamez:
I agree with what @Nyholm said. If StyleCI wouldn't perform the CS checks anyways on each PR, I would perhaps see it differently... But if a developer has to install the PHP-CS-Fixer globally anyways, they can as well install the bridge manually as well.

sagikazarmark:
@jeromegamez upvote means "Let's merge it", while as far as I can see you are against it.

dbu:
i see both points. voted with a confused smiley. not sure which way is better. as its only dev dependencies its not horrible.

jeromegamez:
To be honest, I'm always switching between "include" and "not include" the php-cs-fixer in new projects I start. The thing is: if the developer uses PHP CS Fixer on their machine, the styleci-bridge being available in the dev dependencies is a nice convenience. If they don't, it's an additional library with no use to them. So it's not a strict dev-*requirement* without the PHP-CS-Fixer.

Now I agree with @dbu, I see both points, and I wouldn't oppose either decision ¯\_(ツ)_/¯

sagikazarmark:
You are not making it easier :joy:

Nyholm:
> i see both points. voted with a confused smiley. not sure which way is better. as its only dev dependencies its not horrible.

Some libraries are deadly serious about adding dev dependencies... I do not think we should be too serious but we should really consider the word "dependencies". As I said, our code does not **depend** on the cs fixer... 

To exaggerate and take my argument to an extent: Should we add the [ocular.phar](https://github.com/php-http/client-common/blob/master/.travis.yml#L43) to git because it is a tool  developers can use on our code? Of course not.

sagikazarmark:
Well, I don't insist on this one. One argument that hasn't been said that we could spare some download time. On the other hand, installing globally might not always be a good idea when you need separate versions.

I am looking at phar.io at the moment which might be a solution for tools themselves, but not for libraries like this. Would be nice to have some kind of phar-plugins.

GrahamCampbell:
:-1: on having php-cs-fixer on client side.

sagikazarmark:
@GrahamCampbell Care to elaborate? Do you mean the dependency or the usage of php-cs-fixer itself?

sagikazarmark:
Alright, since there are more people against, I am closing this for now. Thanks @Soullivaneuh anyway.

Soullivaneuh:
@sagikazarmark This was not about having php-cs-fixer on client side, but styleci-bridge package...

Anyway, you have the diff if you want to integrate it.

PR #56, opened 2017-01-20
cast StreamInterface::getSize return in string into ContentLengthPlugin

### LINKED ISSUE #55: ContentLengthPlugin doesn't work with Diactoros
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no

#### Actual Behavior

ContentLengthPlugin has a conflict when Diactoros is used. `MessageInterface::withHeader`  expects a string or a string collection for the `value` parameter and `Diactoros` checks if `value` respects this. ContentLengthPlugin inject an integer so an exception is thrown by Diactoros.

#### Expected Behavior

inject a string for the `value` paramter into `MessageInterface::withHeader` method

#### Steps to Reproduce

use the `ContentLengthPlugin` with `Diactoros`

#### Possible Solutions

cast `StreamInterface::getSize` return value in string [here](https://github.com/php-http/client-common/blob/master/src/Plugin/ContentLengthPlugin.php#L30) : 

```
$request = $request->withHeader('Content-Length', (string) $stream->getSize());
```

lunika:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #55 
| License         | MIT

#### What's in this PR?

Fix issue #55

### DIFF
--- a/spec/Plugin/ContentLengthPluginSpec.php
+++ b/spec/Plugin/ContentLengthPluginSpec.php
@@ -25,7 +25,7 @@ class ContentLengthPluginSpec extends ObjectBehavior
         $request->hasHeader('Content-Length')->shouldBeCalled()->willReturn(false);
         $request->getBody()->shouldBeCalled()->willReturn($stream);
         $stream->getSize()->shouldBeCalled()->willReturn(100);
-        $request->withHeader('Content-Length', 100)->shouldBeCalled()->willReturn($request);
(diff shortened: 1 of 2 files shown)


dbu:
thank you!

sagikazarmark:
Some CHANGELOG entry would be nice to have.

PR #61, opened 2017-03-14
Fix deflate encoding handling and disable compress encoding

ajgarlag:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no/yes
| Deprecations?   | no/yes
| License         | MIT

#### What's in this PR?

This PR updates the DecoderPlugin to use the right decoder to manage deflate content-encoding and disable compress as there is no support in PHP and nobody uses it

This is the solution proposed by @joelwurtz in https://github.com/php-http/message/issues/64#issuecomment-255958592 to the problem is described in https://github.com/php-http/message/issues/64

#### Example Usage

``` php
<?php
require 'vendor/autoload.php';

$decoderPlugin = new Http\Client\Common\Plugin\DecoderPlugin();
$pluginClient = new Http\Client\Common\PluginClient(
    Http\Discovery\HttpClientDiscovery::find(),
    [$decoderPlugin]
);

$response = $pluginClient->sendRequest(
    new GuzzleHttp\Psr7\Request('GET', 'http://httpbin.org/deflate')
);

var_dump($response->getBody()->getContents());

```

The result should be a JSON document like the one you can view at http://httpbin.org/deflate, but it returns a response with an empty body.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Change Log
 
+## Unreleased
+
+### Deprecated
+
+- `DecoderPlugin` does not longer claim to support `compress` content encoding
+
+### Fixed
+
+- `DecoderPlugin` uses the right `FilteredStream` to handle `deflate` content encoding
+
 ## 1.4.1 - 2017-02-20
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


joelwurtz:
no BC break here for me, even if we remove some headers, it just was wrong and was not working so 👍

sagikazarmark:
Thanks

ajgarlag:
Thank you all for you OSS work! A patch release would be useful!

sagikazarmark:
https://github.com/php-http/client-common/releases/tag/v1.4.2

PR #63, opened 2017-03-17
Cross subdomain cookie sharing issue

### LINKED ISSUE #62: Cookies from subdomains are ignored by the CookiePlugin
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 1.4.1

#### Actual Behavior

Cookie coming from the same domain but different subdomain is ignored. 

#### Expected Behavior

The cookie for subdomain should be caught and saved to cookie jar (unless current behavior is what you have initially planned). 
Or, at least we should provide a way for developer to choose if he wants to restrict cookies to single domain. 

#### Steps to Reproduce

Imagine that initial request is made to www.example.com which tries to set cookie for all subdomains: 
`.example.com`

Lines 71-74 will ignore that cookie, since `strpos('.example.com', 'www.example.com')` will return `false`:
```php
// Restrict setting cookie from another domain
if (false === strpos($cookie->getDomain(), $request->getUri()->getHost())) {
    continue;
}
```

#### Possible Solutions

A quick and dirty solution would be to simply switch the arguments inside `strpos()`, like this:
```php
// Restrict setting cookie from another domain
if (false === strpos($request->getUri()->getHost(), $cookie->getDomain())) {
    continue;
}
```

I can make a pull request with that change if you find it appropriate.

However, as I have said earlier, it would be better to let the developer decide whether to accept different domain cookies or not.

andrewkharook:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #62
| License         | MIT

#### What's in this PR?

The changes make cookies set for main domain available for subdomains as well, as described in [RFC 6265](https://tools.ietf.org/html/rfc6265#section-5.2.3)

#### Checklist

### DIFF
--- a/src/Plugin/CookiePlugin.php
+++ b/src/Plugin/CookiePlugin.php
@@ -69,7 +69,7 @@ final class CookiePlugin implements Plugin
                     }
 
                     // Restrict setting cookie from another domain
-                    if (false === strpos($cookie->getDomain(), $request->getUri()->getHost())) {
+                    if (false === strpos($request->getUri()->getHost(), $cookie->getDomain())) {
                         continue;
                     }
 
--- a/src/Plugin/CookiePlugin.php
+++ b/src/Plugin/CookiePlugin.php
@@ -69,7 +69,11 @@ final class CookiePlugin implements Plugin
                     }
 
                     // Restrict setting cookie from another domain
-                    if (false === strpos($request->getUri()->getHost(), $cookie->getDomain())) {
+                    if (false === strpos(
+                            '.'.$request->getUri()->getHost(),
+                            '.'.$cookie->getDomain()
+                        )
+                    ) {
                         continue;
                     }
 
--- a/src/Plugin/CookiePlugin.php
+++ b/src/Plugin/CookiePlugin.php
@@ -69,11 +69,7 @@ final class CookiePlugin implements Plugin
                     }
 
                     // Restrict setting cookie from another domain
-                    if (false === strpos(
-                            '.'.$request->getUri()->getHost(),
-                            '.'.$cookie->getDomain()
-                        )
-                    ) {
+                    if (!preg_match("/\.{$cookie->getDomain()}$/", '.'.$request->getUri()->getHost())) {
                         continue;
                     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 
+- `CookiePlugin` allows main domain cookies to be sent/stored for subdomains
 - `DecoderPlugin` uses the right `FilteredStream` to handle `deflate` content encoding
 
 ## 1.4.1 - 2017-02-20
--- a/spec/Plugin/CookiePluginSpec.php
+++ b/spec/Plugin/CookiePluginSpec.php
@@ -82,6 +82,41 @@ class CookiePluginSpec extends ObjectBehavior
         }, function () {});
     }
 
+    function it_does_not_load_cookie_on_hackish_domains(RequestInterface $request, UriInterface $uri, Promise $promise)
+    {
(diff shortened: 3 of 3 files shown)


src/Plugin/CookiePlugin.php
```diff
@@ -69,7 +69,7 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
                     }
 
                     // Restrict setting cookie from another domain
-                    if (false === strpos($cookie->getDomain(), $request->getUri()->getHost())) {
+                    if (false === strpos($request->getUri()->getHost(), $cookie->getDomain())) {
```
dbu: is this correct. what about symfony.slack.com setting the cookie and then we do a request to hacksymfony.slack.com? its an edge case, but i think we should be correct about this.

sagikazarmark: On the other hand, hack.symfony.slack.com should be correct again. Cookies.... 😕

andrewkharook: @dbu, @sagikazarmark:
What if we add a dot `.` to both cookie domain and request domain before the comparison?
```php
strpos(
    '.'.$request->getUri()->getHost(),
    '.'.$cookie->getDomain()
)
```
Here's what's happening in this case:
```php
strpos(
    '.hacksymfony.slack.com',
    '.symfony.slack.com'
); // no match, cookie ignored.

strpos(
    '.hack.symfony.slack.com',
    '.symfony.slack.com'
); // match, cookie used.
```
I wonder if this workaround is appropriate and not counted as dirty code.

dbu: i think that is the right approach. afaik the server could send the leading `.` already, we should handle that properly as well, not that we end up with "..symfony.slack.com".

are you sure that cookies should propagate down several levels? like test.com cookie being sent to foo.bar.baz.test.com ?

andrewkharook: > afaik the server could send the leading . already, we should handle that properly as well, not that we end up with "..symfony.slack.com".

This is already handled by [Http\Message\Cookie::normalizeDomain()](https://github.com/php-http/message/blob/master/src/Cookie.php#L497), so we shouldn't care about it.

> are you sure that cookies should propagate down several levels? like test.com cookie being sent to foo.bar.baz.test.com ?

Yes, that's what is said in RFC 6265. If Domain attribute is not specified, it defaults to the host portion of the current document location (but not including subdomains). If a domain is specified, subdomains are always included.

src/Plugin/CookiePlugin.php
```diff
@@ -69,7 +69,11 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
                     }
 
                     // Restrict setting cookie from another domain
-                    if (false === strpos($cookie->getDomain(), $request->getUri()->getHost())) {
+                    if (false === strpos(
+                            '.'.$request->getUri()->getHost(),
+                            '.'.$cookie->getDomain()
```
fbourigault: Isn't this check too weak (see https://3v4l.org/nXCMW)? You should here test that `$request->getUri()->getHost()` ends with `$cookie->getDomain()`.

dbu: i think you are right @fbourigault 

@andrewkharook can you please adjust this accordingly and add a test that fails with this code but not with the new code? like comparing fake.domain.com.evil.org with domain.com?

Nyholm:
Thank you @andrewkharook for this PR. Would you mind updating it according to the feedback?

andrewkharook:
Hi, @Nyholm! 
Sure, the update will be ready over the weekend.

Nyholm:
Ping @fbourigault

joelwurtz:
Thanks !

PR #64, opened 2017-03-21
deprecate debug_plugin options

### LINKED ISSUE #60: [RFC] Deprecate PluginClient debug_plugins option
I don't know the whole history of debug_plugins option (which was introduced in #22), but since this option is no longer used by HTTPlugBundle (https://github.com/php-http/HttplugBundle/pull/128), it could be deprecated and then removed in the next major version.

This will keep the PluginClient more simple.

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | fixes #60 
| Documentation   | N/A
| License         | MIT

#### What's in this PR?

The `PluginClient` now triggers a deprecation notice when the `debug_plugins` option is set.

#### Why?

The `debug_plugins` option was used by the HttplugBundle collector, but since https://github.com/php-http/HttplugBundle/pull/128 is merged, it's no longer required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Change Log
 
 
+## Unreleased
+
+### Deprecated
+
+- The `debug_plugins` option for `PluginClient` is deprecated and will be removed in 2.0. Use the decorator design pattern instead like in [ProfilePlugin](https://github.com/php-http/HttplugBundle/blob/de33f9c14252f22093a5ec7d84f17535ab31a384/Collector/ProfilePlugin.php).
+
 ## 1.4.2 - 2017-03-18
 
 ### Deprecated
--- a/src/PluginClient.php
+++ b/src/PluginClient.php
@@ -108,6 +108,10 @@ final class PluginClient implements HttpClient, HttpAsyncClient
      */
     private function configure(array $options = [])
     {
+        if (isset($options['debug_plugins'])) {
+            @trigger_error('The "debug_plugins" option is deprecated since 1.5 and will be removed in 2.0.', E_USER_DEPRECATED);
+        }
+
         $resolver = new OptionsResolver();
         $resolver->setDefaults([
             'max_restarts' => 10,


spec/PluginClientSpec.php
```diff
@@ -87,47 +87,4 @@ function it_throws_loop_exception(HttpClient $httpClient, RequestInterface $requ
 
         $this->shouldThrow('Http\Client\Common\Exception\LoopException')->duringSendRequest($request);
     }
-
-    function it_injects_debug_plugins(HttpClient $httpClient, ResponseInterface $response, RequestInterface $request, Plugin $plugin0, Plugin $plugin1, Plugin $debugPlugin)
```
dbu: should we not continue to test this while its in the codebase?

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # Change Log
 
 
+## Unreleased
+
+### Deprecated
+
+- The `debug_plugins` option for `PluginClient` is deprecated and will be removed in 2.0.
+
```
dbu: would be cool to mention what to use instead

fbourigault: Like "Use the decorator design pattern instead."?

spec/PluginClientSpec.php
fbourigault: I didn't know why I removed this test... We must keep it as the feature is still present and still usable.

CHANGELOG.md
dbu: yep. and maybe point to one place where we do that ourselves as an illustration. thats a bit verbose for a changelog but i would prefer that over writing extensive upgrade guides...

fbourigault:
All requested changes have been addressed.

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # Change Log
 
 
+## Unreleased
+
+### Deprecated
+
+- The `debug_plugins` option for `PluginClient` is deprecated and will be removed in 2.0. Use the decorator design pattern instead like in https://github.com/php-http/HttplugBundle/blob/master/Collector/ProfilePlugin.php.
```
sagikazarmark: Maybe use a markdown link here instead of the whole URL. Also, this is not a permanent link. A link with a commit hash in it might be better.

src/PluginClient.php
```diff
@@ -108,6 +108,10 @@ public function sendAsyncRequest(RequestInterface $request)
      */
     private function configure(array $options = [])
     {
+        if (isset($options['debug_plugins'])) {
+            @trigger_error('The "debug_plugins" option is deprecated and will be removed in 2.0.', E_USER_DEPRECATED);
```
sagikazarmark: I would add `deprecated since X.Y` here as we usually do.

CHANGELOG.md
dbu: when you are on a github page of the file, you can press the `y` key to get the link with the last commit id.

sagikazarmark:
Would be nice to have some docs about this, decorating all plugins manually can be painful an complicated.

Nyholm:
Thank you!

PR #67, opened 2017-03-27
Added QueryDefaultsPlugin

### LINKED ISSUE #66: [RFC] QueryDefaultPlugin
I talk to an API where I identify myself with query parameters. (yes, that is wrong but I cannot change this.) I see a value in having a plugin much like `DefaultHeaderPlugin` but for URI queries.

Nyholm:
This will fix #66

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change Log
 
-## 1.5.0 - Unreleased
+## Unreleased
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change Log
 
+## 1.5.0 - Unreleased
+
+### Added
+
+- `QueryDefaultsPlugin` to add default query parameters. 
 
 ## 1.4.2 - 2017-03-18
 
--- a/spec/Plugin/QueryDefaultsPluginSpec.php
+++ b/spec/Plugin/QueryDefaultsPluginSpec.php
@@ -2,25 +2,26 @@
 
 namespace spec\Http\Client\Common\Plugin;
 
-use PhpSpec\Exception\Example\SkippingException;
+use Http\Client\Common\Plugin;
 use Psr\Http\Message\RequestInterface;
-use Psr\Http\Message\StreamInterface;
 use PhpSpec\ObjectBehavior;
-use Prophecy\Argument;
 use Psr\Http\Message\UriInterface;
 
 class QueryDefaultsPluginSpec extends ObjectBehavior
 {
-    public function it_is_initializable()
+    public function let()
     {
         $this->beConstructedWith([]);
+    }
+
+    public function it_is_initializable()
+    {
         $this->shouldHaveType('Http\Client\Common\Plugin\QueryDefaultsPlugin');
     }
 
     public function it_is_a_plugin()
     {
-        $this->beConstructedWith([]);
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Change Log
 
+## 1.5.0 - Unreleased
```
sagikazarmark: Hm, 1.5 or Unreleased? Using a version there is confusing, that's why we usually avoid it.

spec/Plugin/QueryDefaultsPluginSpec.php
```diff
@@ -0,0 +1,39 @@
...
+use Psr\Http\Message\StreamInterface;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Psr\Http\Message\UriInterface;
+
+class QueryDefaultsPluginSpec extends ObjectBehavior
+{
+    public function it_is_initializable()
+    {
+        $this->beConstructedWith([]);
```
sagikazarmark: you can add this to a `let` method, it's like setUp in phpunit. If you use beConstructedWith again, it will override the on in the let method.

spec/Plugin/QueryDefaultsPluginSpec.php
```diff
@@ -0,0 +1,39 @@
...
+        $this->shouldHaveType('Http\Client\Common\Plugin\QueryDefaultsPlugin');
```
sagikazarmark: I think we use `::class` constants wherever possible. phpspec 3.0 generates specs like that by default.

src/Plugin/QueryDefaultsPlugin.php
```diff
@@ -0,0 +1,53 @@
...
+ */
+final class QueryDefaultsPlugin implements Plugin
+{
+    /**
+     * @var array
+     */
+    private $queryParams = [];
+
+    /**
+     * @param array $queryParams Hashmap of query name to query value
```
dbu: if i read the code correctly, keys and values should *not* be http encoded, right? i think we should mention that here to avoid confusion and double-encoded parameters

Nyholm:
Thank you for your reviews. I have updated the PR accordingly.

spec/Plugin/QueryDefaultsPluginSpec.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public function it_is_initializable()
+    {
+        $this->shouldHaveType('Http\Client\Common\Plugin\QueryDefaultsPlugin');
+    }
+
+    public function it_is_a_plugin()
+    {
+        $this->shouldImplement(Plugin::class);
```
Nyholm: This will break php 5.4 compatibility. Even though I think we should drop 5.4 support, this is not a good enough reason. 

FYI @sagikazarmark

src/Plugin/QueryDefaultsPlugin.php
```diff
@@ -0,0 +1,53 @@
...
+ */
+final class QueryDefaultsPlugin implements Plugin
+{
+    /**
+     * @var array
+     */
+    private $queryParams = [];
+
+    /**
+     * @param array $queryParams Hashmap of query name to query value. Names and values should not be url encoded.
```
dbu: "should" is too weak. i would say "Names and values must not be url encoded as this plugin will encode them"

Nyholm:
Thank you for your reviews. New tag is coming up.

PR #71, opened 2017-05-11
add plugin client factory

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | partially https://github.com/php-http/HttplugBundle/issues/109
| Documentation   | TODO
| License         | MIT

This is required to display plugins created by 3rd party libraries in the HttplugBundle profiler. See https://github.com/php-http/HttplugBundle/issues/109#issuecomment-300461535 for all puzzle pieces.

#### To Do

- [x] Update the changelog.
- [ ] Update the documentation.

### DIFF
--- a/src/PluginClientFactory.php
+++ b/src/PluginClientFactory.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Http\Client\Common;
+
+use Http\Client\HttpAsyncClient;
+use Http\Client\HttpClient;
+
+/**
+ * @author Fabien Bourigault <bourigaultfabien@gmail.com>
+ */
+final class PluginClientFactory
+{
+    /**
+     * @var callable
+     */
+    private static $factory;
+
+    /**
+     * Set the factory to use.
+     * The callable to provide must have the same arguments and return type as PluginClientFactory::createClient.
+     * This is used by the HTTPlugBundle to provide a better Symfony integration.
+     *
+     * @internal
+     *
+     * @param callable $factory
+     */
+    public static function setFactory(callable $factory)
+    {
+        static::$factory = $factory;
+    }
+
+    /**
+     * @param HttpClient|HttpAsyncClient $client
+     * @param Plugin[]                   $plugins
+     * @param array                      $options {
+     *
+     *     @var string $client_name to give client a name which may be used when displaying client information  like in
+     *         the HTTPlugBundle profiler.
+     * }
+     *
+     * @see PluginClient constructor for PluginClient specific $options.
+     *
+     * @return PluginClient
+     */
+    public function createClient($client, array $plugins = [], array $options = [])
+    {
+        if (static::$factory) {
+            $factory = static::$factory;
+
+            return $factory($client, $plugins, $options);
+        }
+
+        return new PluginClient($client, $plugins, $options);
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+- `PluginClientFactory` to create `PluginClient` instances.
+
 ### Deprecated
 
 - The `debug_plugins` option for `PluginClient` is deprecated and will be removed in 2.0. Use the decorator design pattern instead like in [ProfilePlugin](https://github.com/php-http/HttplugBundle/blob/de33f9c14252f22093a5ec7d84f17535ab31a384/Collector/ProfilePlugin.php).
--- a/spec/PluginClientFactorySpec.php
+++ b/spec/PluginClientFactorySpec.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace spec\Http\Client\Common;
+
+use Http\Client\HttpClient;
+use PhpSpec\ObjectBehavior;
+
+class PluginClientFactorySpec extends ObjectBehavior
+{
+    function it_is_initializable()
+    {
+        $this->shouldHaveType('Http\Client\Common\PluginClientFactory');
+    }
+
+    function it_returns_a_plugin_client(HttpClient $httpClient)
+    {
+        $client = $this->createClient($httpClient);
+
+        $client->shouldHaveType('Http\Client\Common\PluginClient');
+    }
+}


src/PluginClientFactory.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Http\Client\Common;
+
+/**
+ * Create a PluginClient.
```
Nyholm: We could have a more expressive comment.

src/PluginClientFactoryInterface.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace Http\Client\Common;
+
+use Http\Client\HttpAsyncClient;
+use Http\Client\HttpClient;
+
+/**
+ * Create a PluginClient.
```
Nyholm: Same here. :)

src/PluginClientFactoryInterface.php
```diff
@@ -0,0 +1,25 @@
...
+ * @author Fabien Bourigault <bourigaultfabien@gmail.com>
+ */
+interface PluginClientFactoryInterface
+{
+    /**
+     * @param HttpClient|HttpAsyncClient $client
+     * @param Plugin[]                   $plugins
+     * @param array                      $options
+     *
+     * @see PluginClient constructor for $options details.
```
Nyholm: No, we cannot "see the PluginClient". We are not aware of any client implementation. We should write something like "array of options that are passed to the client."

src/PluginClientFactory.php
Nyholm: Or maybe remove this class comment and only use the one on the interface.

src/PluginClientFactoryInterface.php
fbourigault: I don't think so. There is only one PluginClient without a dedicated interface and this implementation is final. There is no way to write an other PluginClient!

src/PluginClientFactoryInterface.php
fbourigault: I'm open to any suggestion :)

src/PluginClientFactoryInterface.php
Nyholm: You are correct. I had in mind that one could create ANY `PluginClient` with this. But the `PluginClientFactoryInterface` creates an instance of the concrete class `PluginClient`. That is a huge different from our other `*FactoryInterface`s. They create an instance of an interface.

src/PluginClientFactoryInterface.php
Nyholm: Maybe: 

> Create a PluginClient object loaded with an array of plugins. A factory may add or modify the plugins or change the HttpClient.

Nyholm:
Note: 
The `PluginClientFactoryInterface` creates an instance of the concrete class `PluginClient`. That is a huge different from our other `*FactoryInterface`s. They create an instance of an interface.

sagikazarmark:
TBH, I don't really see the point of this interface.

fbourigault:
The idea behind is to use discovery to find a concrete PluginClientFactory. When the bundle profiling is enabled, a profile aware PluginClientFactory is decorates each plugin with a ProfilePlugin and add the StackPlugin at the begining. See https://github.com/php-http/HttplugBundle/issues/109#issuecomment-300461535 for a whole plan details.

sagikazarmark:
Why do we need a factory to decorate plugins? And why do we have to discover it? I still think it's configuration: when profiling is enabled, hook into DI and wrap plugins, otherwise don't.

fbourigault:
It's all about displaying plugins from 3rd party like php-github-api in the profiler. With such implementation, you can integrate any 3rd party library (which use this interface and the discovery) in the profiler without having any specific code in the 3rd party library, without having a glue bundle and without having to write a line of configuration.

fbourigault:
I updated the PR to add a note in the changelog. I also adressed @Nyholm comments about docblocks. I added in my to do list the puli stuff. Can someone help me with it?

fbourigault:
How can I test puli resources declaration?

fbourigault:
This is now updated to provide a solution which does not require `php-http/discovery` to profile 3rd party library clients.

src/PluginClientFactory.php
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * @var callable
+     */
+    private static $factory;
+
+    /**
+     * Set the factory to use. The callable to provide must have the same arguments and return type as
+     * PluginClientFactory::createClient.
+     *
+     * @internal
```
Nyholm: Maybe state why?

> This is used by the HTTPlugBundle to provide a better Symfony integration

src/PluginClientFactory.php
```diff
@@ -0,0 +1,48 @@
...
+     * @param Plugin[]                   $plugins
+     * @param array                      $options
+     *
+     * @see PluginClient constructor for $options details.
+     *
+     * @return PluginClient
+     */
+    public function createClient($client, array $plugins = [], array $options = [])
+    {
+        $factory = static::$factory ?: function ($client, array $plugins, array $options) {
```
Nyholm: Why not simply do: 

```php
if (static::$factory) {
  return $factory($client, $plugins, $options);
}

return new PluginClient($client, $plugins, $options);
```

It would be easier to read.

fbourigault:
Thank you @Nyholm. I addressed all your remarks.

src/PluginClientFactory.php
```diff
@@ -0,0 +1,49 @@
...
+     * @param array                      $options
+     *
+     * @see PluginClient constructor for $options details.
+     *
+     * @return PluginClient
+     */
+    public function createClient($client, array $plugins = [], array $options = [])
+    {
+        if (static::$factory) {
+            return (static::$factory)($client, $plugins, $options);
```
Nyholm: This is not allowed in PHP 5.x.  ( I just found out)

```php
$func = static::$factory;
return $func($client, $plugins, $options);

// Or: (This might work)
return static::$factory($client, $plugins, $options); 
```

fbourigault:
Should be fixed now. Please, don't merge until it's ok on other packages.

Nyholm:
@fbourigault Is this ready to merge? 

Do @dbu or @sagikazarmark want to have a final review?

src/PluginClientFactory.php
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * @author Fabien Bourigault <bourigaultfabien@gmail.com>
+ */
+final class PluginClientFactory
+{
+    /**
+     * @var callable
+     */
+    private static $factory;
```
sagikazarmark: Is this the workaround for not having an interface? If so, having an interface for this is still less evil. This solution provides zero type safety (even if the support PHP versions do not support return types).

fbourigault: As the stuff is marked internal, we choosed to avoid over engineering and to not use an interface.

fbourigault:
I added the `client_name` option.

Nyholm:
I will merge this, but we should review it a final time before releasing. 

Thank you for the comments and thank you for the work on this PR

src/PluginClientFactory.php
```diff
@@ -0,0 +1,55 @@
...
+     * @see PluginClient constructor for PluginClient specific $options.
+     *
+     * @return PluginClient
+     */
+    public function createClient($client, array $plugins = [], array $options = [])
+    {
+        if (static::$factory) {
+            $factory = static::$factory;
+
+            return $factory($client, $plugins, $options);
```
dbu: deos simply `return static::$factory(...)` not work?

src/PluginClientFactory.php
```diff
@@ -0,0 +1,55 @@
...
+     *
+     *     @var string $client_name to give client a name which may be used when displaying client information  like in
+     *         the HTTPlugBundle profiler.
+     * }
+     *
+     * @see PluginClient constructor for PluginClient specific $options.
+     *
+     * @return PluginClient
+     */
+    public function createClient($client, array $plugins = [], array $options = [])
```
dbu: why is this method dynamic but the factory callable static? this means i must instantiate the factory, but i can only set one single callable. why not make the callable an optional constructor argument?

src/PluginClientFactory.php
fbourigault: Only in PHP 7+

dbu:
could we use the class docblock to explain *why* this factory should be used?

src/PluginClientFactory.php
fbourigault: The setFactory is static as we need to change the implementation at runtime when bundle profiling is enabled and, as we aim to do 0config integration, we can't have something non-static here.

src/PluginClientFactory.php
dbu: weird. can you add a comment that this is to work with php 5 and can be removed in 7?

src/PluginClientFactory.php
dbu: ah, i see. maybe a bit of comment in the code would help so we remember in the future ;-)

src/PluginClientFactory.php
fbourigault: I was wrong. It doesn't work in any PHP version: "Function name must be a string".
See https://3v4l.org/WA5ek.

dbu: alright. then maybe a comment that using it directly does not work at least up to php 7.1?

fbourigault: Doesn't work too in 7.2.0alpha2. IMHO it doesn't worth to write a comment about something not supported by the language itself.

PR #74, opened 2017-06-06
Add AddPathPlugin always_prepend option

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| Documentation   | TODO
| License         | MIT

This allow to make the `AddPathPlugin` smart. By smart I mean to not prepend the configured URI path to the request URI if already present. This behavior is disabled by default to keep code BC.

The `BaseUriPlugin` now has a third optional constructor argument to configure the underlying `AddPathPlugin`.

This is required for https://github.com/m4tthumphrey/php-gitlab-api which has a base URI set to `https://example.gitlab.com/api/v3` but is also used to follow pagination `Link` headers uri which already have the base URI configured.

## To do
* [ ] Documentation.
* [x] Changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Added
+
+- Add the `always_prepend` option to `AddPathPlugin`.
+- The `BaseUriPlugin` can now passthru options for the underlying `AddPathPlugin`.
+
 ### Deprecated
 
 - The `debug_plugins` option for `PluginClient` is deprecated and will be removed in 2.0. Use the decorator design pattern instead like in [ProfilePlugin](https://github.com/php-http/HttplugBundle/blob/de33f9c14252f22093a5ec7d84f17535ab31a384/Collector/ProfilePlugin.php).
--- a/src/Plugin/AddPathPlugin.php
+++ b/src/Plugin/AddPathPlugin.php
@@ -5,6 +5,7 @@ namespace Http\Client\Common\Plugin;
 use Http\Client\Common\Plugin;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\UriInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
  * Prepend a base path to the request URI. Useful for base API URLs like http://domain.com/api.
@@ -19,9 +20,18 @@ final class AddPathPlugin implements Plugin
     private $uri;
 
     /**
-     * @param UriInterface $uri
+     * @var bool
      */
-    public function __construct(UriInterface $uri)
+    private $alwaysPrepend;
+
+    /**
+     * @param UriInterface $uri
+     * @param array        $config {
+     *
+     *     @var bool $alwaysPrepend Set to true to always prepend the path even if the request path start with that path.
+     * }
+     */
+    public function __construct(UriInterface $uri, array $config = [])
     {
         if ($uri->getPath() === '') {
             throw new \LogicException('URI path cannot be empty');
@@ -32,6 +42,12 @@ final class AddPathPlugin implements Plugin
         }
 
         $this->uri = $uri;
+
+        $resolver = new OptionsResolver();
+        $this->configureOptions($resolver);
+        $options = $resolver->resolve($config);
+
+        $this->alwaysPrepend = $options['always_prepend'];
     }
 
     /**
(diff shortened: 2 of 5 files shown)


fbourigault:
I'm open to suggestions for an alternative to the smart name!

src/Plugin/AddPathPlugin.php
```diff
@@ -19,9 +20,18 @@
...
 
     /**
+     * @var bool
+     */
+    private $smart;
+
+    /**
      * @param UriInterface $uri
+     * @param array        $config {
+     *
```
dbu: please remove the extra blank line

src/Plugin/AddPathPlugin.php
```diff
@@ -19,9 +20,18 @@
...
+     *     @var bool $smart True will prepend path only if request path does not start with path to add.
```
dbu: i am somehow confused by the wording. maybe "Set to true to only prepend the path only if the request path does not start with that path already."?

src/Plugin/AddPathPlugin.php
```diff
@@ -32,17 +42,39 @@ public function __construct(UriInterface $uri)
...
      */
     public function handleRequest(RequestInterface $request, callable $next, callable $first)
     {
-        $request = $request->withUri($request->getUri()
-            ->withPath($this->uri->getPath().$request->getUri()->getPath())
-        );
+        $prepend = $this->uri->getPath();
+        $path = $request->getUri()->getPath();
+
+        if (!($this->smart && substr($path, 0, strlen($prepend)) === $prepend)) {
```
dbu: can we rewrite this to `if (!$this->smart || !substr(...))`? i feel that is more readable.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Added
+
+- Add the `$smart` option to `AddPathPlugin` to prepend path only when required.
```
dbu: lets write this `'smart'` instead of `$smart`

fbourigault:
About the _smart_ name. We can also reverse the boolean meaning and so the behavior (and of course change the default value). Maybe it could help to find a better name.

src/Plugin/AddPathPlugin.php
fbourigault: Looks like StyleCI wants this extra blank line.

Nyholm:
> what about always_prepend with true as default value?

I like that better.

src/Plugin/AddPathPlugin.php
```diff
@@ -32,17 +41,39 @@ public function __construct(UriInterface $uri)
...
      */
     public function handleRequest(RequestInterface $request, callable $next, callable $first)
     {
-        $request = $request->withUri($request->getUri()
-            ->withPath($this->uri->getPath().$request->getUri()->getPath())
-        );
+        $prepend = $this->uri->getPath();
+        $path = $request->getUri()->getPath();
+
+        if (!$this->smart || substr($path, 0, strlen($prepend)) !== $prepend) {
```
Nyholm: This only consider the case where `$prepend` is the beginning of `$path`. Is that the only case we want? 

In your case on Gitlab, you do not want to prepend if a absolute URL is used. IMHO, that logic should only be in BaseUriPlugin.

fbourigault: To get back to my use case, it's about using `Link` header href attributes. As of https://tools.ietf.org/html/rfc5988 §5.1 this IRI can be relative so I'm not sure that in a future release gitlab will send relative IRIs.

I did this way to be more consistant with `AddHostPlugin` which owns the replace logic.

fbourigault: But maybe I'm wrong with the client design and the API path (/api/v3) should be hardcoded in the abstract api class and so, I just need to use the `AddHostPlugin` as the user will only configure it's gitlab instance domain.

sagikazarmark:
How about `preserve_uri` defaulting to `false`?

src/Plugin/AddPathPlugin.php
Nyholm: > But maybe I'm wrong with the client design and the API path (/api/v3) should be hardcoded in the abstract api class 

Yeah, maybe.. :/ 

Or in every endpoint.. 


Though. This PR is still useful.

sagikazarmark:
Hm, rereading the initial comment not sure what we prepend here. Always prepend sounds a little weird to me. `prepend/preserve`_`path/uri`?

Nyholm:
The AddPath plugin prepends the current path with a predefined value. 

The idea is that if the predefined value is "/foo/bar" then we should not prepend that to an url like "/foo/bar/whatever" since it already contains "/foo/bar".

sagikazarmark:
No, my issue is with `always`. Despite the word, it does not express the conditions when it matters. Of course we can document it properly, but very likely the next time I meet the config value I will just not remember what it was and look it up again.

Don't know if there is a better name for it though.

Nyholm:
Something between “always_prepend” and “prepend_even_if_uri_start_with_prefix” would be good. 

Regards
Tobias Nyholm

sagikazarmark:
Exactly 😂

fbourigault:
*Fabien listen to propositions.*

Nyholm:
Name it `prepend_even_if_uri_start_with_prefix`

fbourigault:
Closing because no longer required for https://github.com/m4tthumphrey/php-gitlab-api.

mxr576:
Can we re-open this issue? If you try to restart a failed API call by calling `$first($request)` in a plugin (ex.: when an Oauth access token expires and you want your API client to automatically resend the same request with a new access provided by a custom Oauth auth plugin) then the current behaviour of AddPathPlugin corrupts the URL by adding the API version prefix to the URL again.

mxr576:
What is the exact use case behind adding a path prefix to a path that already has the prefix anyway? I can not imagine any.

PR #75, opened 2017-06-06
Use trusty dist for travis-ci build

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| Documentation   | N/A
| License         | MIT

This is about fixing travis-ci HHVM build. See https://travis-ci.org/php-http/client-common/jobs/240025931.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: php
 
 sudo: false
 
+dist: trusty
+
 cache:
     directories:
         - $HOME/.composer/cache/files


.travis.yml
```diff
@@ -2,6 +2,8 @@ language: php
 
 sudo: false
 
+dist: trusty
```
Nyholm: Let's run trusty only on HHVM. Or does it matter?

fbourigault: Everything is green. So it probably does not matter :P

Nyholm:
Thanks!

Issue #77, opened 2017-06-28
[RFC] Error plugin separate 400 / 500


joelwurtz:
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes

#### Actual Behavior

Actually the error plugin always throws an exception on status >= 400

#### Expected Behavior

It would be better to add options for only catching status >= 500 or 500 > status >= 400

As an example when i want to retry requests on a web server, i may want to retrieve only on server errors and not on client errors.

Also 400 errors are often not errors and more a normal comportement that i need to take into account, as i would like to know if some resource exists on the server i don't want to deal that with an exception having a 404 may be a normal thing for me.

However 500 errors are never normals so i want to treat them specialy compared to the 400 ones.

WDYT ?

Nyholm:
Sure, Let's add some config to the error plugin.

PR #79, opened 2017-07-01
Allow a callable parameter to decide if require is going to be retrie…

### LINKED ISSUE #78: Retry plugin backoff
Would be useful if the retry plugin would backoff before retrying.

Nyholm:
…d or not.

Also allow retry backoff

This will fix #78 

It is a minor breaking change here. Instead of retry immediately we wait a second. I do not consider it a BC break though..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 ## Unreleased
 
+### Added
+
+- Added new option 'delay' for `RetryPlugin`.
+- Added new option 'decider' for `RetryPlugin`.
+
+### Changed
+
+- The `RetryPlugin` does now wait between retries. To disable/change this feature you must write something like: 
+ 
+```php
+$plugin = new RetryPlugin(['delay' => function(RequestInterface $request, Exception $e, $retries) { 
+  return 0; 
+}); 
+```
+
 ### Deprecated
 
 - The `debug_plugins` option for `PluginClient` is deprecated and will be removed in 2.0. Use the decorator design pattern instead like in [ProfilePlugin](https://github.com/php-http/HttplugBundle/blob/de33f9c14252f22093a5ec7d84f17535ab31a384/Collector/ProfilePlugin.php).
--- a/spec/Plugin/RetryPluginSpec.php
+++ b/spec/Plugin/RetryPluginSpec.php
@@ -102,7 +102,7 @@ class RetryPluginSpec extends ObjectBehavior
         $this->handleRequest($request, $next, function () {})->shouldReturnAnInstanceOf('Http\Client\Promise\HttpFulfilledPromise');
     }
 
-    function it_has_a_good_default_delay(RequestInterface $request, Exception\HttpException $exception)
+    function it_has_an_exponential_default_delay(RequestInterface $request, Exception\HttpException $exception)
     {
         $this->defaultDelay($request, $exception, 0)->shouldBe(1000);
         $this->defaultDelay($request, $exception, 1)->shouldBe(2000);
--- a/spec/Plugin/RetryPluginSpec.php
+++ b/spec/Plugin/RetryPluginSpec.php
@@ -101,4 +101,12 @@ class RetryPluginSpec extends ObjectBehavior
         $this->handleRequest($request, $next, function () {})->shouldReturnAnInstanceOf('Http\Client\Promise\HttpFulfilledPromise');
         $this->handleRequest($request, $next, function () {})->shouldReturnAnInstanceOf('Http\Client\Promise\HttpFulfilledPromise');
     }
+
+    function it_has_a_good_default_delay(RequestInterface $request, Exception\HttpException $exception)
+    {
+        $this->defaultDelay($request, $exception, 0)->shouldBe(1000);
+        $this->defaultDelay($request, $exception, 1)->shouldBe(2000);
+        $this->defaultDelay($request, $exception, 2)->shouldBe(4000);
+        $this->defaultDelay($request, $exception, 3)->shouldBe(8000);
+    }
 }
--- a/src/Plugin/RetryPlugin.php
+++ b/src/Plugin/RetryPlugin.php
@@ -57,7 +57,7 @@ final class RetryPlugin implements Plugin
             'decider' => function (RequestInterface $request, Exception $e) {
                 return true;
             },
-            'delay' =>  __CLASS__.'::defaultDelay',
+            'delay' => __CLASS__.'::defaultDelay',
         ]);
         $resolver->setAllowedTypes('retries', 'int');
         $resolver->setAllowedTypes('decider', 'callable');
@@ -101,7 +101,7 @@ final class RetryPlugin implements Plugin
             usleep($time);
 
             // Retry in synchrone
-            $this->retryStorage[$chainIdentifier]++;
+            ++$this->retryStorage[$chainIdentifier];
             $promise = $this->handleRequest($request, $next, $first);
 
             return $promise->wait();
@@ -110,8 +110,8 @@ final class RetryPlugin implements Plugin
 
     /**
      * @param RequestInterface $request
-     * @param Exception $e
-     * @param int $retries The number of retries we made before. First time this get called it will be 0.
+     * @param Exception        $e
+     * @param int              $retries The number of retries we made before. First time this get called it will be 0.
      *
      * @return int
      */
--- a/src/Plugin/RetryPlugin.php
+++ b/src/Plugin/RetryPlugin.php
@@ -45,8 +45,8 @@ final class RetryPlugin implements Plugin
      * @param array $config {
      *
      *     @var int $retries Number of retries to attempt if an exception occurs before letting the exception bubble up.
-     *     @var callable $decider A callback that gets a request and an exception to decide if we should retry this or not.
-     *     @var callable $delay A callback to return how many milliseconds we should wait before trying again.
+     *     @var callable $decider A callback that gets a request and an exception to decide after a failure whether the request should be retried.
+     *     @var callable $delay A callback that gets a request, an exception and the number of retries and returns how many milliseconds we should wait before trying again.
      * }
      */
     public function __construct(array $config = [])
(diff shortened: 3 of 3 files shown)


src/Plugin/RetryPlugin.php
```diff
@@ -35,18 +45,30 @@
      * @param array $config {
      *
      *     @var int $retries Number of retries to attempt if an exception occurs before letting the exception bubble up.
+     *     @var callable $decider A callback that gets a request and an exception to decide if we should retry this or not.
```
dbu: maybe: ... to decide after a failure whether the request should be retried.

src/Plugin/RetryPlugin.php
```diff
@@ -35,18 +45,30 @@
...
+     *     @var callable $delay A callback to return how many milliseconds we should wait before trying again.
```
dbu: we should document the arguments for this callback as well.

src/Plugin/RetryPlugin.php
```diff
@@ -73,7 +95,12 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
                 throw $exception;
             }
 
-            ++$this->retryStorage[$chainIdentifier];
+            if (!call_user_func($this->decider, $request, $exception)) {
+                throw $exception;
+            }
+
+            $time = call_user_func($this->delay, $request, $exception, ++$this->retryStorage[$chainIdentifier]);
+            usleep($time);
```
dbu: this freezes the whole php process, even when doing async http requests.

src/Plugin/RetryPlugin.php
```diff
@@ -35,18 +45,30 @@
...
+     *     @var callable $delay A callback to return how many milliseconds we should wait before trying again.
      * }
      */
     public function __construct(array $config = [])
     {
         $resolver = new OptionsResolver();
         $resolver->setDefaults([
             'retries' => 1,
+            'decider' => function (RequestInterface $request, Exception $e) {
+                return true;
```
dbu: should we check the exception class and only do this when the exception indicates a network problem or a server error? 4xx responses or completely invalid requests don't need to be retried. (though we could want to retry dns errors)

if we make the logic more complicated, we should move this to a static public method of this class instead of the anonymous function, also to allow testing it.

src/Plugin/RetryPlugin.php
```diff
@@ -35,18 +45,30 @@
...
+            },
+            'delay' => function (RequestInterface $request, Exception $e, $retries) {
+                return ((int) pow(2, $retries - 1)) * 1000;
```
dbu: this could be a public static function on this class. then we could unit test it.

src/Plugin/RetryPlugin.php
Nyholm: I do not think we can add **any** logic here because it will break the existing behavior.

dbu: i agree. we could argue that its more of a bugfix because retrying something that can't ever work out is pointless. but its quite possible we miss some situation where retrying could be successful, and thus we better change to something like that when we do a version 2. should we add an issue for that? or a todo comment in the code?

Nyholm:
> when doing multiple requests that fail, we will unconditionally delay for each of the requests. i wonder if we could instead work with a timestamp. like recording that response was received at time X and if we get here we wait until X+$time. for a single request, this will result in the same delay, but if we sent out a bunch of requests we will only sleep once if we follow that logic.

Does this makes sense? Or I mean, does it work?

If one do two async requests that both will fail. At some point request A will fail, we check the timestamp and we sleep. Request B will not fail until Request A has been retried, right? 
When Request B fails we take the new timestamp..

It would only make sense if we slept between: 

```php
$promise = $this->handleRequest($request, $next, $first);
return $promise->wait();
```

Or am I wrong?

dbu:
> Request B will not fail until Request A has been retried, right?

i guess you are correct. can we find a timestamp coming from the curl library that was doing the actual async requests? because those should happen in actual threads... if we can't i guess this has to be good enough.

Nyholm:
We can't I'm afraid.. Since we do not know what client they are using.

fbourigault:
> It would only make sense if we slept between:
```php
$promise = $this->handleRequest($request, $next, $first);
return $promise->wait();
```
You have to sleep before `$promise->wait()` but you can also sleep before `$promise = $this->handleRequest($request, $next, $first);`. Or you can do something like:
```
$time = microtime(true);
$promise = $this->handleRequest($request, $next, $first);
usleep($delay + ($time - microtime(true)) * 1000);
return $promise->wait();
```
This will remove the `$promise = $this->handleRequest($request, $next, $first);` execution time from the delay to sleep. It can save a really few µs.

dbu:
i don't think saving microseconds is worth it. its an error case and we wait for a second or similar, which is magnitudes longer. the only problem i see is multiple waits of 1 (or several) seconds each.

src/Plugin/RetryPlugin.php
GrahamCampbell: Technically, adding sleeps is not a semantic change. How can program tell if it has been switched out by the scheduler? It is technically possible, but extremely unlikely, for the original code to have also slept there, if it was not allocated time.

Nyholm:
I do not think there is a solution to the problem you highlight. I updated the PR with a change log. I think this is ready to merge

src/Plugin/RetryPlugin.php
joelwurtz: I don't think we should check the exception here, 4XX can have sens to be rety in some scenario (like you have a pool of servers and you know that some files are present on some of them so you want to retry until you find the file)

That's why i propose the change in #77 so an user can choose on which behavior he wants exception (only 500, only 400 or both of them)

Then if an user choose only 500, 4XX response will never have exception and enter this callback, thus having a check here is not needed

joelwurtz:
For the problem of having to wait here i recently come to an idea to be full asynchronous here instead of using the wait, but it will require some codes :).

First we need a custom Promise implementation which we cannot resolve and we will return this promise in this plugin :

````php
$promise = new CustomPromiseImplementation();

...

return $promise;
```

Then when we receive the response from the upstream promise, we can resolve our custom promise with response if it's OK

```php
$promise = new CustomPromiseImplementation();

$next($request)->then(function (ResponseInterface $response) use ($request, $chainIdentifier) {
             if (array_key_exists($chainIdentifier, $this->retryStorage)) {
                 unset($this->retryStorage[$chainIdentifier]);
             }

             $promise->resolve($response);
 
             return $response;
         }
...

return $promise;
```

And then when it fails we can resend a new async request and resolve or make it failed (when too many retries) our custom promise just like before.

If nobody is up to do that i may throw a PR for this in the summer to resolve this but i have so many things to do right now that not sure if i will have the time for it so feel free to do it.

joelwurtz:
Also using something like mean we can reuse this principle in other code and drop the wait method from our interface

src/Plugin/RetryPlugin.php
dbu: @joelwurtz we agreed to keep it like this for BC reasons. can you please put your argument into #80 so that it does not get lost?

dbu:
interesting idea, joel. i suggest we merge this anyways and see for other solutions. if there will be a BC break involved, we can do a 2.0 version. is that ok for you?

sagikazarmark:
@joelwurtz I remember you wanted to use async-interop. Is this something that they do as well. If not, what's your current point of view on this question: should we still lean towards async-interop?

CHANGELOG.md
```diff
@@ -2,6 +2,21 @@
...
 ## Unreleased
 
+### Added
+
+- Added new option 'delay' for `RetryPlugin`.
+- Added new option 'decider' for `RetryPlugin`.
+
+### Changed
+
+- The `RetryPlugin` does now wait between retries. To disable/change this feature you must write something like:
```
sagikazarmark: now waits

spec/Plugin/RetryPluginSpec.php
```diff
@@ -101,4 +101,12 @@ function it_does_not_keep_history_of_old_failure(RequestInterface $request, Resp
         $this->handleRequest($request, $next, function () {})->shouldReturnAnInstanceOf('Http\Client\Promise\HttpFulfilledPromise');
         $this->handleRequest($request, $next, function () {})->shouldReturnAnInstanceOf('Http\Client\Promise\HttpFulfilledPromise');
     }
+
+    function it_has_a_good_default_delay(RequestInterface $request, Exception\HttpException $exception)
```
sagikazarmark: I'm sorry, but what does good mean? 😂 Dat software specification 😂

src/Plugin/RetryPlugin.php
```diff
@@ -73,12 +93,30 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
+    /**
+     * @param RequestInterface $request
+     * @param Exception        $e
+     * @param int              $retries The number of retries we made before. First time this get called it will be 0.
+     *
+     * @return int
+     */
+    public static function defaultDelay(RequestInterface $request, Exception $e, $retries)
+    {
+        return ((int) pow(2, $retries)) * 1000;
```
sagikazarmark: I guess `$retries` can only be a positive integer or 0, so why the int casting?

CHANGELOG.md
Nyholm: Are you sure? 'It does now waits'?

spec/Plugin/RetryPluginSpec.php
Nyholm: Haha. You know... not "bad". 
I'll update :)

src/Plugin/RetryPlugin.php
Nyholm: Not sure to be honest. Just copied this from guzzle.

Nyholm:
I've updated the PR

joelwurtz:
Let's go merge this, my comment is not a blocker, and can be done later.

@sagikazarmark async-interop is dead so don't know actually, i would like to changing the async interface to be used with coroutines instead (like async / await in other languages, see php amp for the details), but may be a bit soon for the php community.

bendavies:
@Nyholm just started using this. the default `defaultDelay` doesn't seem like a great default
with 5 retries, the plugin retries pretty much instantly from 0.001 to 0.032 seconds. no delay at all really. Clearly I can define my own, delay, but maybe the default could be a bit more ...delaying?

Nyholm:
Hm, I guess you are correct. We should update our defaults. Could you make a suggestion in a PR?

bendavies:
sure.

Issue #82, opened 2017-07-06
Content-type plugin


kpn13:
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | 

Hi,

Do you think a plugin which will automatically set the content-type header based on what it can find in the body request would be a good idea ? I can do a pull request on it if you approved this idea.

Thanks in advance.

joelwurtz:
Hey, it can be nice yes, wondering how you will do it as from my experience it's far from being easy :)

kpn13:
Hum I don't say it will be easy :). Well, if I start by detect json content and xml content it is I think relatively simple and will respond to major needs, no ?

joelwurtz:
Yes would be a good start, but implementation will matters, as i will certainly be against a too complex one that is hard to maintain.

PR #85, opened 2017-07-21
Add Content-type plugin to set content-type header automatically

kpn13:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | mentioned in #82 
| Documentation   | https://github.com/php-http/documentation/pull/204
| License         | MIT

#### What's in this PR?

Add a new plugin which allow to detect JSON and XML content in body request and automatically set the right content-type header.

#### Why?

Set the content-type header is often unpleasing and detect whether the content is JSON or XML is relatively simple so it could save time.

#### Example Usage

``` php
$contentTypePlugin = new ContentTypePlugin();

$pluginClient = new PluginClient(
    HttpClientDiscovery::find(),
    [$contentTypePlugin]
);
```

#### Checklist

- [x ] Documentation pull request created (if not simply a bugfix)

### DIFF
--- a/spec/Plugin/ContentTypePluginSpec.php
+++ b/spec/Plugin/ContentTypePluginSpec.php
@@ -56,4 +56,13 @@ class ContentTypePluginSpec extends ObjectBehavior
         $this->handleRequest($request, function () {}, function () {});
     }
 
+    function it_does_not_set_content_type_header_if_size_0_or_unknown(RequestInterface $request)
+    {
+        $request->hasHeader('Content-Type')->shouldBeCalled()->willReturn(false);
+        $request->getBody()->shouldBeCalled()->willReturn(\GuzzleHttp\Psr7\stream_for());
+        $request->withHeader('Content-Type', null)->shouldNotBeCalled();
+
+        $this->handleRequest($request, function () {}, function () {});
+    }
+
 }
--- a/spec/Plugin/ContentTypePluginSpec.php
+++ b/spec/Plugin/ContentTypePluginSpec.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace spec\Http\Client\Common\Plugin;
+
+use PhpSpec\Exception\Example\SkippingException;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\StreamInterface;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+
+class ContentTypePluginSpec extends ObjectBehavior
+{
+    function it_is_initializable()
+    {
+        $this->shouldHaveType('Http\Client\Common\Plugin\ContentTypePlugin');
+    }
+
+    function it_is_a_plugin()
+    {
+        $this->shouldImplement('Http\Client\Common\Plugin');
+    }
+
+    function it_adds_json_content_type_header(RequestInterface $request)
+    {
+        $request->hasHeader('Content-Type')->shouldBeCalled()->willReturn(false);
+        $request->getBody()->shouldBeCalled()->willReturn(json_encode(['foo' => 'bar']));
+        $request->withHeader('Content-Type', 'application/json')->shouldBeCalled()->willReturn($request);
+
+        $this->handleRequest($request, function () {}, function () {});
+    }
+
+    function it_adds_xml_content_type_header(RequestInterface $request)
+    {
+        $request->hasHeader('Content-Type')->shouldBeCalled()->willReturn(false);
+        $request->getBody()->shouldBeCalled()->willReturn('<foo>bar</foo>');
+        $request->withHeader('Content-Type', 'application/xml')->shouldBeCalled()->willReturn($request);
+
+        $this->handleRequest($request, function () {}, function () {});
+    }
+
+    function it_does_not_set_content_type_header(RequestInterface $request)
+    {
+        $request->hasHeader('Content-Type')->shouldBeCalled()->willReturn(false);
+        $request->getBody()->shouldBeCalled()->willReturn('foo');
+        $request->withHeader('Content-Type', null)->shouldNotBeCalled();
+
+        $this->handleRequest($request, function () {}, function () {});
+    }
+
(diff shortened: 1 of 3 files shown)


spec/Plugin/ContentTypePluginSpec.php
```diff
@@ -0,0 +1,59 @@
...
+        $request->getBody()->shouldBeCalled()->willReturn(json_encode(['foo' => 'bar']));
+        $request->withHeader('Content-Type', 'application/json')->shouldBeCalled()->willReturn($request);
+
+        $this->handleRequest($request, function () {}, function () {});
+    }
+
+    function it_adds_xml_content_type_header(RequestInterface $request)
+    {
+        $request->hasHeader('Content-Type')->shouldBeCalled()->willReturn(false);
+        $request->getBody()->shouldBeCalled()->willReturn('<foo>bar</foo>');
```
joelwurtz: Can you use a real stream implementation instead of a string ?

I think this test case will fail with a stream implementation as there is no rewind

src/Plugin/ContentTypePlugin.php
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * @param $stream
+     *
+     * @return bool
+     */
+    private function isJson($stream)
+    {
+        json_decode($stream);
+
```
joelwurtz: We should rewind the stream here

src/Plugin/ContentTypePlugin.php
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * @param $stream
+     *
+     * @return \SimpleXMLElement|false
+     */
+    private function isXml($stream)
+    {
+        libxml_use_internal_errors(true);
+
```
joelwurtz: We should rewind the stream here

src/Plugin/ContentTypePlugin.php
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @author Karim Pinchon <karim.pinchon@gmail.com>
+ */
+final class ContentTypePlugin implements Plugin
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
```
joelwurtz: We should avoid content type detection if stream is not rewindable

src/Plugin/ContentTypePlugin.php
```diff
@@ -0,0 +1,62 @@
...
+
+use Http\Client\Common\Plugin;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Allow to set the correct content type header on the request automatically only if it is not set .
+ *
+ * @author Karim Pinchon <karim.pinchon@gmail.com>
+ */
+final class ContentTypePlugin implements Plugin
```
joelwurtz: May be nice to add some options to this plugin to avoid detection on too large stream, or stream with a null size

src/Plugin/ContentTypePlugin.php
```diff
@@ -0,0 +1,62 @@
...
+                $request = $request->withHeader('Content-Type', 'application/json');
+
+                return $next($request);
+            }
+
+            if ($this->isXml($stream)) {
+                $request = $request->withHeader('Content-Type', 'application/xml');
+
+                return $next($request);
+            }
```
joelwurtz: Some content type that can be added to check (but not necessarly in this PR) :

 * Image types by using a library image (should be optional IMO)
 * Binary type by checking if there is a null character (\0) in the stream (that's how git does to detect binary vs text)
 * Text type (inverse of the binary type, imo the default content type we want to add to the stream)

src/Plugin/ContentTypePlugin.php
```diff
@@ -0,0 +1,62 @@
...
+ * @author Karim Pinchon <karim.pinchon@gmail.com>
+ */
+final class ContentTypePlugin implements Plugin
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
+        if (!$request->hasHeader('Content-Type')) {
```
joelwurtz: We should early return here if the body size of the stream is equal to 0

src/Plugin/ContentTypePlugin.php
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /**
+     * @param $stream
+     *
+     * @return \SimpleXMLElement|false
+     */
+    private function isXml($stream)
+    {
+        libxml_use_internal_errors(true);
```
Nyholm: Should we reset to the original value?

src/Plugin/ContentTypePlugin.php
```diff
@@ -0,0 +1,123 @@
...
+
+            if (!$stream->isSeekable()) {
+                return $next($request);
+            }
+
+            if (0 == $streamSize) {
+                return $next($request);
+            }
+
+            if ($this->skipDetection && $streamSize >= $this->sizeLimit) {
```
Nyholm: What if $streamSize == null?

src/Plugin/ContentTypePlugin.php
```diff
@@ -0,0 +1,123 @@
...
+    {
+        if (!$request->hasHeader('Content-Type')) {
+            $stream = $request->getBody();
+            $streamSize = $stream->getSize();
+
+            if (!$stream->isSeekable()) {
+                return $next($request);
+            }
+
+            if (0 == $streamSize) {
```
Nyholm: We should use strict comparison and then do an extra one for checking if null.

joelwurtz: Not here, as we skip every stream that has a size equal to null, don't we want to skip only thoses with skipDetection and null size ?

I mean if stream is rewindable, but we cannot determine it's size but still want to auto detect the content type why skipping this ?

kpn13: I agree, I made it too quickly.

PR #88, opened 2017-11-10
Allow Symfony 4

OndraM:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

#### What's in this PR?
Allow Symfony/OptionsResolver 4.x installation.

#### Why?
Symfony 4 is about to be released this month (and is already in beta). There are no BC breaks in the OptionsResolver component: https://github.com/symfony/options-resolver/compare/3.4...master , so I think we can safely allow its installation.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         "php-http/httplug": "^1.1",
         "php-http/message-factory": "^1.0",
         "php-http/message": "^1.6",
-        "symfony/options-resolver": "^2.6 || ^3.0"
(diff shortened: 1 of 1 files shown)


makasim:
any ETA? It block me from upgrading enqueue/stomp lib and others.

Nyholm:
This is awesome. I will make sure this is merged during the Symfony con hack day. I will also fix all other php-http packages.

Nyholm:
Thank you.

PR #90, opened 2017-11-18
Test against symfony versions

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | if this is a new feature, link to pull request in https://github.com/php-http/documentation that adds relevant documentation
| License         | MIT

#### What's in this PR?

I want to test with different versions of the options-resolver. 

#### Why?

Im experimenting to see if the symfony/lts is something that should be used in travis.yml for more bundles. 

#SymfonyConHackday2017

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,9 +28,9 @@ matrix:
 
           # Test LTS versions
         - php: 7.1
-          env: DEPENDENCIES="symfony/lts:2"
+          env: DEPENDENCIES="symfony/lts:v2"
         - php: 7.1
-          env: DEPENDENCIES="symfony/lts:3"
+          env: DEPENDENCIES="symfony/lts:v3"
 
           # Latest beta releases
         - php: 7.1
@@ -44,7 +44,7 @@ matrix:
 before_install:
     - if [[ $COVERAGE != true ]]; then phpenv config-rm xdebug.ini || true; fi
     - if [ "$DEPENDENCIES" = "minimun" ]; then COMPOSER_FLAGS="--prefer-stable --prefer-lowest"; fi;
-    - if [ "$DEPENDENCIES" = "beta" ]; then COMPOSER_FLAGS="--stability beta"; fi;
+    - if [ "$DEPENDENCIES" = "beta" ]; then composer config minimum-stability beta; fi;
     - if [ ! -z "$DEPENDENCIES" ]; then composer require --no-update $DEPENDENCIES; fi;
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,9 +19,8 @@ branches:
 matrix:
     fast_finish: true
     include:
-          # Minimum supported PHP and Symfony version
-        - php: 7.1
-          env: DEPENDENCIES="minimum" COVERAGE=true TEST_COMMAND="composer test-ci"
+        - php: 5.5
+          env: DEPENDENCIES="minimum" COVERAGE=true TEST_COMMAND="composer test-ci"  DEPENDENCIES="henrikbjorn/phpspec-code-coverage^1.0"
 
           # Test the latest stable release
         - php: 5.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ matrix:
     fast_finish: true
     include:
         - php: 7.1
-          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" DEPENDENCIES="doctrine/instanciator:^1.05"
+          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" DEPENDENCIES="doctrine/instanciator:^1.0.5"
 
           # Test the latest stable release
         - php: 5.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ matrix:
 before_install:
     - if [[ $COVERAGE != true ]]; then phpenv config-rm xdebug.ini || true; fi
     - if [ "$DEPENDENCIES" = "minimun" ]; then COMPOSER_FLAGS="--prefer-stable --prefer-lowest"; fi;
-    - if [ "$DEPENDENCIES" = "beta" ]; then COMPOSER_FLAGS="--minimum-stability=beta"; fi;
+    - if [ "$DEPENDENCIES" = "beta" ]; then COMPOSER_FLAGS="--stability beta"; fi;
 
 install:
     # To be removed when this issue will be resolved: https://github.com/composer/composer/issues/5355
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ matrix:
         - php: 7.1
           env: DEPENDENCIES="dunglas/symfony-lock:^3"
         - php: 7.1
-          env: DEPENDENCIES="dunglas/symfony-lock:^4"
+          env: DEPENDENCIES="dunglas/symfony-lock:^4" STABILITY="rc"
 
           # Latest dev release
         - php: 7.1
@@ -53,6 +53,7 @@ before_install:
     - if ! [ -z "$DEPENDENCIES" ]; then composer require --no-update ${DEPENDENCIES}; fi;
 
 install:
+    - cat composer.json
     # To be removed when this issue will be resolved: https://github.com/composer/composer/issues/5355
     - if [[ "$COMPOSER_FLAGS" == *"--prefer-lowest"* ]]; then composer update --prefer-dist --no-interaction --prefer-stable --quiet; fi
     - composer update ${COMPOSER_FLAGS} --prefer-dist --no-interaction
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,31 +20,34 @@ matrix:
     fast_finish: true
     include:
           # Minimum supported PHP and Symfony version
-        - php: 5.4
+        - php: 7.1
           env: DEPENDENCIES="minimum" COVERAGE=true TEST_COMMAND="composer test-ci"
 
           # Test the latest stable release
+        - php: 5.6
+        - php: 7.0
         - php: 7.1
+        - php: 7.2
 
           # Test LTS versions
-        - php: 7.1
+        - php: 5.4
           env: DEPENDENCIES="symfony/lts:v2"
-        - php: 7.1
+        - php: 5.5
           env: DEPENDENCIES="symfony/lts:v3"
 
-          # Latest beta release
+          # Latest dev release
         - php: 7.1
-          env: DEPENDENCIES="beta"
+          env: DEPENDENCIES="dev"
 
     allow_failures:
           # Latest beta is allowed to fail.
         - php: 7.1
-          env: DEPENDENCIES="beta"
+          env: DEPENDENCIES="dev"
           
 before_install:
     - if [[ $COVERAGE != true ]]; then phpenv config-rm xdebug.ini || true; fi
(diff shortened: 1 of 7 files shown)


Nyholm:
We would like to keep tests for beta and the 2 latest Symfony LTS. 

Ping @weaverryan.

This does not quite work.. When Symfony 4.0 is released we do no longer test SF3.4... I will do another attempt later.

Nyholm:
I think I found the "perfect" travis.yml. 

The current state of the PR has `"symfony/options-resolver": "^2.6 || ^3.0"`. Travis will test against the following: 
- SF 2.6 (lowest)
- SF 3.3 (Current latest stable)
- SF 2.8 (LTS v2)
- SF 3.3 (LTS v3) (Same as above)
- SF 3.4-beta (Beta) (Allowed to fail)

When Symfony 4 is released and I do not change my library requirements, Travis will test the following: 
- SF 2.6 (lowest)
- SF 3.4 (Current latest stable) <--- Change
- SF 2.8 (LTS v2)
- SF 3.4 (LTS v3) (Same as above) <--- Change
- SF 3.4-beta (Beta) (Allowed to fail) <--- Now meaningless. 

If I update my requirements to `"symfony/options-resolver": "^2.6 || ^3.0 || ^4.0"` after Symfony 4 is released: 
- SF 2.6 (lowest)
- SF 4.0 (Current latest stable) <--- Change
- SF 2.8 (LTS v2)
- SF 3.4 (LTS v3) 
- SF 4.0-beta (Beta) (Allowed to fail) <--- Will have a purpose before 4.0 is released

FYI @stof, @nicolas-grekas

.travis.yml
```diff
@@ -27,11 +19,33 @@ branches:
 matrix:
     fast_finish: true
     include:
+          # Minimum supported PHP and Symfony version
         - php: 5.4
-          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" COVERAGE=true TEST_COMMAND="composer test-ci"
-
+          env: DEPENDENCIES="minimun" COVERAGE=true TEST_COMMAND="composer test-ci"
```
nicolas-grekas: minimum

.travis.yml
```diff
@@ -27,11 +19,33 @@ branches:
...
+
+          # Test the latest stable release
+        - php: 7.1
+
+          # Test LTS versions
+        - php: 7.1
+          env: DEPENDENCIES="symfony/lts:v2"
```
nicolas-grekas: clever use of the package, great discovery!

Nyholm:
Ping @alcaeus

.travis.yml
```diff
@@ -27,11 +19,33 @@ branches:
...
         - php: 5.4
-          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" COVERAGE=true TEST_COMMAND="composer test-ci"
-
+          env: DEPENDENCIES="minimum" COVERAGE=true TEST_COMMAND="composer test-ci"
+
+          # Test the latest stable release
+        - php: 7.1
+
+          # Test LTS versions
+        - php: 7.1
```
alcaeus: Food for thought: test the LTS releases against the lowest common version they and this package supports. I'd say that people on LTS releases are less likely to always run on the latest PHP version, so testing against a lower version might be more comparable to real-world usage.

.travis.yml
```diff
@@ -27,11 +19,33 @@ branches:
...
+          env: DEPENDENCIES="symfony/lts:v3"
+
+          # Latest beta release
+        - php: 7.1
+          env: DEPENDENCIES="beta"
+
+    allow_failures:
+          # Latest beta is allowed to fail.
+        - php: 7.1
+          env: DEPENDENCIES="beta"
```
weaverryan: I think the latest beta should not be allowed to fail. Probably we really should allow it to fail, but I'm afraid people will not "notice" the failure. I'd rather push them to be a bit more proactive by noticing it :)

.travis.yml
weaverryan: What if some bundle authors actively only want to support 2 Symfony versions at one time? What's what we're talking about in FOSUserBundle https://github.com/FriendsOfSymfony/FOSUserBundle/pull/2639#issuecomment-344907215

fbourigault: This can still install a mixed set of 2.7 and 2.8 components. What about using `symfony/symfony:v2.8.*` which will force any symfony component to be in it's v2.8 version?

weaverryan: Sorry, I realize I was unclear! What I *really* mean was: in FOSUserBundle, they only want to support Symfony 3 and Symfony 4 (and then perhaps just do bug-fixes fro 2.7/2.8 branches).

In that situation, I guess they would just remove this *one* line and change the minimum dependencies matrix to 5.5 (from 5.4)?

stof: remove the Symfony 2 LTS job and bump your min requirement

.travis.yml
```diff
@@ -27,11 +19,33 @@ branches:
...
+          # Test the latest stable release
+        - php: 7.1
+
+          # Test LTS versions
+        - php: 7.1
+          env: DEPENDENCIES="symfony/lts:v2"
+        - php: 7.1
+          env: DEPENDENCIES="symfony/lts:v3"
+
+          # Latest beta release
```
stof: I suggest testing against dev rather than beta (especially if you allow them to fail):

- if something breaks in an early Symfony 4.1 version, you may have 5 months to report it or adapt, instead of a few weeks/days
- most of your other deps don't release beta at all, so you will not catch anything until it reaches other jobs too

.travis.yml
dbu: re requiring symfony/symfony: that hides problems if we miss depending on some symfony components. if we do that, we should require each component we use in the configured version instead. (which duplicates composer.json, so still not very nice)

fbourigault: We still have the "Test the latest stable release" job which does not alter the `composer.json`.

.travis.yml
```diff
@@ -27,18 +19,44 @@ branches:
 matrix:
     fast_finish: true
     include:
+          # Minimum supported PHP and Symfony version
```
stof: `Minimum supported PHP` is confusing here

.travis.yml
```diff
@@ -27,18 +19,44 @@ branches:
...
+          env: DEPENDENCIES="symfony/lts:^2"
+        - php: 5.5
+          env: DEPENDENCIES="symfony/lts:^3"
+
+          # Latest dev release
+        - php: 7.1
+          env: DEPENDENCIES="dev"
 
+    allow_failures:
+          # Latest beta is allowed to fail.
```
stof: wrong comment

.travis.yml
```diff
@@ -27,18 +19,44 @@ branches:
...
+        - php: 7.1
```
stof: no need to specify the PHP version here. It will avoid having to keep it in sync when changing the version used by `DEPENDENCIES="dev"`

.travis.yml
```diff
@@ -27,18 +19,44 @@ branches:
 matrix:
     fast_finish: true
     include:
+          # Minimum supported PHP and Symfony version
+        - php: 7.1
+          env: DEPENDENCIES="minimum" COVERAGE=true TEST_COMMAND="composer test-ci"
```
stof: should we actually run the coverage on the min deps ?

src/Plugin/ContentTypePlugin.php
```diff
@@ -102,7 +102,7 @@ private function isJson($stream)
 
         json_decode($stream->getContents());
 
-        return json_last_error() == JSON_ERROR_NONE;
+        return JSON_ERROR_NONE == json_last_error();
```
stof: strict comparison ?

.travis.yml
dbu: i'd do coverage on the latest stable

.travis.yml
```diff
@@ -27,18 +19,44 @@ branches:
...
+          env: DEPENDENCIES="dev"
 
+    allow_failures:
+          # Latest beta is allowed to fail.
+        - php: 7.1
+          env: DEPENDENCIES="dev"
+          
 before_install:
     - if [[ $COVERAGE != true ]]; then phpenv config-rm xdebug.ini || true; fi
+    - if [ "$DEPENDENCIES" = "minimum" ]; then COMPOSER_FLAGS="--prefer-stable --prefer-lowest"; fi;
```
xabbuh: I would instead just set `$COMPOSER_FLAGS` above

.travis.yml
```diff
@@ -27,18 +19,44 @@ branches:
...
+    - if [ "$DEPENDENCIES" = "dev" ]; then composer config minimum-stability dev; fi;
```
xabbuh: I would rename the var to `$STABILITY` (or something like that) here

.travis.yml
```diff
@@ -27,18 +19,44 @@ branches:
...
+    - if [[ $DEPENDENCIES == *"/"* ]]; then composer require --no-update $DEPENDENCIES; fi;
```
xabbuh: and then just test for `$DEPENDENCIES` not being empty here

Nyholm:
Thank you for the review. An Im sorry for keeping you review the actual travis file here and the "best practice" in the docs. 

I've made the updates now

.travis.yml
```diff
@@ -27,18 +19,46 @@ branches:
 matrix:
     fast_finish: true
     include:
+        - php: 7.1
+          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" DEPENDENCIES="doctrine/instanciator:^1.05"
```
fbourigault: It's `1.0.5` ;)

.travis.yml
```diff
@@ -27,18 +19,46 @@ branches:
 matrix:
     fast_finish: true
     include:
+        - php: 7.1
+          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" DEPENDENCIES="doctrine/instanciator:^1.0.5"
```
alcaeus: It's also `instantiator`. Apart from that, any reason why you're locking the version of it? If anything breaks with `--prefer-lowest` because an older version is installed, it needs to be fixed in that package and should probably be added here as a comment so the workaround can be removed.

dbu: this is adding the dependency for this one case, so its changing composer.json and thus needs a version constraint.

alcaeus: I can see that, the question is: why do you need it as an explicit dependency?

.travis.yml
```diff
@@ -27,18 +19,47 @@ branches:
 matrix:
     fast_finish: true
     include:
+        - php: 7.0
+          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" DEPENDENCIES="doctrine/instantiator:^1.1"
```
fbourigault: `doctrine/instantiator:^1.1` require PHP 7.1.

.travis.yml
Nyholm: From Slack: 

> @fbourigault [11:18 AM] 
$ composer why doctrine/instantiator
phpspec/phpspec   2.4.0  requires  doctrine/instantiator (^1.0.1)
phpspec/prophecy  1.4.0  requires  doctrine/instantiator (^1.0.2) (edited)
But it’s for unmaintained versions of those packagesages

Nyholm:
Thank you for all the reviews!

.travis.yml
alcaeus: I see. Thank you for explaining 👍

PR #94, opened 2018-01-04
Improved performance of QueryDefaultsPlugin

vudaltsov:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #67 
| Documentation   | 
| License         | MIT

### DIFF
--- a/spec/Plugin/QueryDefaultsPluginSpec.php
+++ b/spec/Plugin/QueryDefaultsPluginSpec.php
@@ -39,4 +39,21 @@ class QueryDefaultsPluginSpec extends ObjectBehavior
         }, function () {
         });
     }
+
+    public function it_does_not_replace_existing_request_value(RequestInterface $request, UriInterface $uri)
+    {
+        $this->beConstructedWith([
+            'foo' => 'fooDefault',
+            'bar' => 'barDefault',
+        ]);
+
+        $request->getUri()->shouldBeCalled()->willReturn($uri);
+        $uri->getQuery()->shouldBeCalled()->willReturn('foo=new');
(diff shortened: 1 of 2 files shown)


vudaltsov:
@dbu , yes.

`$requestQuery += $defaultQuery` will only add elements from __default query__ not existing in __request query__ (including the ones with numeric indexes) according to http://php.net/manual/en/language.operators.array.php, which is exactly the same as iterating over __default query__ and adding to __request query__ with `!isset` check.

The order will be preserved. The new array will contain elements from `$requestQuery` in their current order with added nonexisting elements from `$defaultQuery` in their original order. Same as it was before.

vudaltsov:
Could this be merged? Or can I help somehow? Maybe add more tests?

Nyholm:
Im not sure about this. A side effect (that I think I see):

```php
$defaults = ['foo'=>'bar'];
$query = ['foo' => 'newValue'];

// Running the plugin

$output = 'foo=bar&foo=newValue';
```

Right?

vudaltsov:
No, the output will be `$output = 'foo=newValue'`. We parse the given query and merge default values into it (with existing values ignored). And then encode the clean query array and pass it to request.

My solution should give no side effects.

Nyholm:
okey. Maybe I misread the code. Is there a test that cover this case?

vudaltsov:
Sure! I hope I have time this weekend to commit such a test.

vudaltsov:
done

vudaltsov:
@Nyholm , let's merge it?

joelwurtz:
Thanks for this @vudaltsov and sorry for the wait, yes let's merge this

Issue #98, opened 2018-03-13
FilteredStream: unable to read content of stream multiple times


mxr576:
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 1.7.0

I am working on a 3rd party integration for my library that uses HTTplug and this library. When I integrated my library to Drupal 8 that has its own Guzzle client wrapper I faced with an interesting problem. By using the `\Http\Client\Common\Plugin\DecoderPlugin` sometimes I got back an empty string instead of the JSON that API has actually returned.

I think this is the combination of two possible issues:

1.  If something reads the content of the full response body before the `\Http\Client\Common\Plugin\DecoderPlugin` plugin can do that then when it starts processing the response (ex.: with the `\Http\Message\Encoding\DechunkStream`) it does not rewinds the stream, therefore when the `\Http\Message\Encoding\FilteredStream::getContents()` is being called the while loop does not do anything because the wrapped (underlying) stream has already in the end.
Maybe this can be solved by adding this condition to the constructor of the FilteredStream class.

```php
if ($this->stream->isSeekable()) {
  $this->stream->rewind();
}
```

2. If the wrapped stream was being closed then any of the `\Http\Message\Encoding\FilteredStream` implementations could not read the content of the stream again. It means that if I use the `\Http\Client\Common\Plugin\DecoderPlugin` on my client an try to get the body of the response multiple times like this: `(string) $response->getBody()` then after the first call I always get back an empty string. (Or if we add the suggested condition to the FilteredStream class' constructor an exception, because the read() method tries to read on a closed stream after the first `(string) $response->getBody()`). If I remove the plugin then the originally wrapped `\GuzzleHttp\Psr7\Stream`  object inside an instance of `\GuzzleHttp\Psr7\Response` can always return the same response body (string) multiple times.

mxr576:
In my opinion a possible solution for the 2nd point could be storing only a copy of the `$stream` parameter inside an instance of `FilteredStream` class by using the `MessageFactoryDiscovery` to find a suitable class that can "deep copy" a stream. That would introduce a new dependency to the `php-http/discovery` in this library.

mxr576:
[...]

PR #99, opened 2018-03-14
Add buffered stream to decoder plugin

joelwurtz:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #98 
| License         | MIT

#### What's in this PR?

Add a buffer option to decoder plugin

#### Why?

Allow to seek / rewind a stream which has been filtered

#### To Do

- [ ] Add tests
- [ ] Add options to pass to the buffered stream (memory max usage, buffer type : file / memory)

### DIFF
--- a/src/Plugin/DecoderPlugin.php
+++ b/src/Plugin/DecoderPlugin.php
@@ -4,6 +4,7 @@ namespace Http\Client\Common\Plugin;
 
 use Http\Client\Common\Plugin;
 use Http\Message\Encoding;
+use Http\Message\Stream\BufferedStream;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\StreamInterface;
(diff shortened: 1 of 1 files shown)


src/Plugin/DecoderPlugin.php
```diff
@@ -123,18 +133,28 @@ private function decodeOnEncodingHeader($headerName, ResponseInterface $response
...
-            return new Encoding\GzipDecodeStream($stream);
+            $decorated = new Encoding\GzipDecodeStream($stream);
+        }
+        
+        if ($decorated === null) {
+            return null;   
+        }
+        
+        if ($this->buffered) {
+            $decorated = new BufferedStream($decorated);
```
joelwurtz: In fact this is a bad place as we only need this buffer once at the end, if we have a stream that is both gzip and chunked encoded we will have 2 buffered streams :/

joelwurtz:
close in favor of #167

PR #100, opened 2018-04-05
Add an option to only throw exception on 5XX error codes

joelwurtz:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #77 
| Documentation   | https://github.com/php-http/documentation/pull/235/
| License         | MIT

cf #77

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## Upcoming
+
+### Added
+
+ - Add an option on ErrorPlugin to only throw exception on response with 5XX status code.
+
 ## 1.7.0 - 2017-11-30
 
 ### Added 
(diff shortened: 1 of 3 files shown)


joelwurtz:
> I like this feature. Do you think it is a good idea (or overkill) to let the user provide a regex to decide when to throw an exception?

I think it's overkill, someone can still do it's own plugin for that, however handling 4XX responses is common (like you want the bad request for handling validations, or 404 so you know that this resource does not exist and can handle it properly, but 5XX responses are never wanted because client don't know how to respond to them)

sagikazarmark:
Like the idea. Is this PR ready?

joelwurtz:
For me yes

sagikazarmark:
Great. Can you write some documentation as well?

CHANGELOG.md
```diff
@@ -1,11 +1,15 @@
 # Change Log
 
-## Upcoming
+## 1.8 (unreleased)
```
GrahamCampbell: 1.8.0?

dbu: i guess we can still add the .0 when we do the release

PR #101, opened 2018-04-05
Use deferred to be full async in retry

joelwurtz:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

#### What's in this PR?

It adds a deferred class which allow to resolve retry in an async way (no more wait call)

#### Why?

Previously using retry plugin would force async to be sync

#### Checklist

- [x] Need to updated CHANGELOG.md (wait for other PR to be merged to not have useless conflicts)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Changed
 
 - AddPathPlugin no longer add prefix multiple times if a request is restarted - it now only adds the prefix if that request chain has not yet passed through the AddPathPlugin
+- RetryPlugin no longer wait for retried requests and use a deferred promise instead
 
 ## 1.7.0 - 2017-11-30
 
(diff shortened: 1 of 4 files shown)


dbu:
this can now be rebased and made ready for merging ;-)

joelwurtz:
It should be good now.

However this plugin may still need a warning, as we have a call to the sleep function inside which is synchronous, there is the possibility to have asynchronous sleep but it heavily depends on the implementation (amp / react / ...), so there is no possibility to add it here

dbu:
> However this plugin may still need a warning, as we have a call to the sleep function inside which is synchronous

do you want to add something for this topic to the documentation of the retry plugin?

joelwurtz:
Done in https://github.com/php-http/client-common/pull/101

PR #104, opened 2018-05-08
Fix Cookie handling as per RFC 6265 Section 5.4

### LINKED ISSUE #54: CookiePlugin adds multiple Cookie headers to the Request
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | v1.4.0, and probably previous as well

#### Actual Behavior

When using the CookiePlugin with a CookieJar filled with multiple applicable Cookies, each will add a new `Cookie: key=value` header to the Request. Or maybe concatenate them, separated by a command `, `.

#### Expected Behavior

Only one Cookie header should be added to the request. Where key-value pair should be separated by `; `.

[RFC 6265 Section 5.4](https://tools.ietf.org/html/rfc6265#section-5.4):
>   When the user agent generates an HTTP request, the user agent MUST
   NOT attach more than one Cookie header field.

#### Steps to Reproduce

```php
$cookieJar = new \Http\Message\CookieJar();
$cookieJar->addCookies([
    new \Http\Message\Cookie('foo', 'bar'),
    new \Http\Message\Cookie('cookie', 'oreo'),
]);

$plugin = new \Http\Client\Common\Plugin\CookiePlugin($cookieJar);

$request = new \GuzzleHttp\Psr7\Request('GET', 'http://127.0.0.1/');
$response = new \GuzzleHttp\Psr7\Response();

$noop = function(\GuzzleHttp\Psr7\Request $request) use($response) {
    var_dump($request->getHeaderLine('Cookie'));
    return new Http\Promise\FulfilledPromise($response);
};

$plugin->handleRequest($request, $noop, $noop);

// "foo=bar, cookie=oreo"
```

It's a little hard to demonstrate with only the `client-common` package, but when using the HttplugBundle one can easily see the problem when firing a request at [httpbin](https://httpbin.org)/[RequestBin](https://requestb.in)

(I was using the Guzzle6 client btw)

#### Possible Solutions

I think the problem lies in the fact the `Request::withAddedHeader('Cookie', ...)` gets called multiple times, as can seen [here](https://github.com/php-http/client-common/blob/3cf7eb93a2e19bded055efff1a267a1fa6d46074/src/Plugin/CookiePlugin.php#L56).

The `\GuzzleHttp\Psr7\Request` object doesn't handle the Cookie header in any special way, so it won't _join_ those lines the proper way.

I believe the CookieJar should join the cookie key-value pairs and only set the Cookie header once.

**Some feedback would be welcome, as I'm not 100% sure where this issue should be resolved.**  (In this package, in guzzle/psr7 or the client(s))

---

If the route of fixing this package would be taken, you might also want to consider implementing step 2 mentioned in [RFC 6265 Section 5.4](https://tools.ietf.org/html/rfc6265#section-5.4):

> 2.  The user agent SHOULD sort the cookie-list in the following
>    order:
> 
>    *  Cookies with longer paths are listed before cookies with
>       shorter paths.
> 
>    *  Among cookies that have equal-length path fields, cookies with
>       earlier creation-times are listed before cookies with later
>       creation-times.

gnat42:
Concatenate multiple cookies

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #54 
| License         | MIT

#### What's in this PR?
Makes the cookie plugin only add one cookie header with all cookies separated by ;

#### Checklist

#### To Do

- [x] Need to add tests - I couldn't figure out the phpspec syntax, where the RequestInterface was being created etc. But I'm willing to add if someone can give me some pseudo code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## 1.8.2 (unreleased)
+
+### Changed
+
+- When multiple cookies exist, a single header with all cookies is sent as per RFC 6265 Section 5.4
+
 ## 1.8.1 - 2018-10-09
 
 ### Fixed
--- a/spec/Plugin/CookiePluginSpec.php
+++ b/spec/Plugin/CookiePluginSpec.php
@@ -51,6 +51,27 @@ class CookiePluginSpec extends ObjectBehavior
         }, function () {});
     }
 
+    function it_combines_multiple_cookies_into_one_header(RequestInterface $request, UriInterface $uri, Promise $promise)
+    {
+        $cookie = new Cookie('name', 'value', 86400, 'test.com');
+        $cookie2 = new Cookie('name2', 'value2', 86400, 'test.com');
+
+        $this->cookieJar->addCookie($cookie);
+        $this->cookieJar->addCookie($cookie2);
+
+        $request->getUri()->willReturn($uri);
+        $uri->getHost()->willReturn('test.com');
+        $uri->getPath()->willReturn('/');
+
+        $request->withAddedHeader('Cookie', 'name=value; name2=value2')->willReturn($request);
+
+        $this->handleRequest($request, function (RequestInterface $requestReceived) use ($request, $promise) {
+            if (Argument::is($requestReceived)->scoreArgument($request->getWrappedObject())) {
+                return $promise->getWrappedObject();
+            }
+        }, function () {});
+    }
+
     function it_does_not_load_cookie_if_expired(RequestInterface $request, UriInterface $uri, Promise $promise)
     {
         $cookie = new Cookie('name', 'value', null, 'test.com', false, false, null, (new \DateTime())->modify('-1 day'));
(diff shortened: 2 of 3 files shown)


gnat42:
I should note this is a similar solution to guzzle's [CookieJar](https://github.com/guzzle/guzzle/blob/master/src/Cookie/CookieJar.php#L293)

joelwurtz:
Totally approve here, for the reference, the specification say that: 

> the user agent MUST NOT attach more than one Cookie header field.

Would you like help for adding test ? Or if you don't have time i can do them. But former solution would be nice, and we can help you better in our [Slack channel](http://slack.httplug.io)

dbu:
Joel, do you have time to add the test so we can merge this fix?

joelwurtz:
No time ATM, sorry for the delay :/

gnat42:
I have no idea why I thought it was so complicated the first time. Writing a test was quite easy. I must have been sleep deprived the first time. It'd be great to get this fixed, merged and released.

gnat42:
I think the latest test failure is unrelated to my change..

dbu:
thanks a lot! that failure looks like a temporary network problem. merged your change.

PR #110, opened 2018-09-25
Fix deferred when in a non async context

joelwurtz:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #109
| License         | MIT

Should fix #109 

Do not have time to make a proper test ATM, if urgent create a ticket for the test, merge this and push a new version

For the explanation it's because fullfilled / rejected call directly the callback before returning a new one, which makes the deferred be resolved before we set the then callback (and so the new deferred will never be resolved / rejected)

PR #111, opened 2018-09-27
Do not use defer

### LINKED ISSUE #109: can only throw objects
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 1.8.0

#### Actual Behavior

Error: Can only throw objects
/home/travis/build/php-http/HttplugBundle/vendor/php-http/client-common/src/Deferred.php:129
/home/travis/build/php-http/HttplugBundle/vendor/php-http/client-common/src/PluginClient.php:87
/home/travis/build/php-http/HttplugBundle/Tests/Functional/ProfilingTest.php:96

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #109
| Documentation   | 
| License         | MIT

I revert the changes to RetryPlugin as a temporary fix for #109. Only consider this if we cannot properly fix it (as attempted in #110)

### DIFF
--- a/spec/Plugin/RetryPluginSpec.php
+++ b/spec/Plugin/RetryPluginSpec.php
@@ -30,9 +30,7 @@ class RetryPluginSpec extends ObjectBehavior
             }
         };
 
-        $promise = $this->handleRequest($request, $next, function () {});
-        $promise->shouldReturnAnInstanceOf('Http\Client\Common\Deferred');
-        $promise->wait()->shouldReturn($response);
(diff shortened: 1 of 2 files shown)


joelwurtz:
Unfortunately i have very little time ATM to deal with the Defer problem :s, so yes, if nobody came up with a solution reverting it is better ATM and will try to tackle the problem latter

PR #113, opened 2018-10-26
Path prefix must be present

mxr576:
Make sure the path prefix is actually present even if the request has been altered before.

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #103
| Documentation   | -
| License         | MIT

#### What's in this PR?

Follow up on #103.

Everything worked fine until I refactored my tests and started to use a singleton API client for offline testing. If I call the same API endpoints in the same test with the client then the first API request gets prefixed properly but the second one does not, because the generated `$identifier` is already exist in the ` $this->alteredRequests`.

### DIFF
--- a/src/Plugin/AddPathPlugin.php
+++ b/src/Plugin/AddPathPlugin.php
@@ -48,7 +48,7 @@ final class AddPathPlugin implements Plugin
     {
         $identifier = spl_object_hash((object) $first);
 
-        if (!array_key_exists($identifier, $this->alteredRequests) || strpos($request->getUri()->getPath(), $this->uri->getPath()) !== 0) {
+        if (!array_key_exists($identifier, $this->alteredRequests) || 0 !== strpos($request->getUri()->getPath(), $this->uri->getPath())) {
             $request = $request->withUri($request->getUri()
                 ->withPath($this->uri->getPath().$request->getUri()->getPath())
             );
--- a/src/Plugin/AddPathPlugin.php
+++ b/src/Plugin/AddPathPlugin.php
@@ -48,7 +48,7 @@ final class AddPathPlugin implements Plugin
     {
         $identifier = spl_object_hash((object) $first);
 
-        if (!array_key_exists($identifier, $this->alteredRequests)) {
+        if (!array_key_exists($identifier, $this->alteredRequests) || strpos($request->getUri()->getPath(), $this->uri->getPath()) !== 0) {
             $request = $request->withUri($request->getUri()
                 ->withPath($this->uri->getPath().$request->getUri()->getPath())
             );


src/Plugin/AddPathPlugin.php
```diff
@@ -48,7 +48,7 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
     {
         $identifier = spl_object_hash((object) $first);
 
-        if (!array_key_exists($identifier, $this->alteredRequests)) {
+        if (!array_key_exists($identifier, $this->alteredRequests) || 0 !== strpos($request->getUri()->getPath(), $this->uri->getPath())) {
```
dbu: should we assign $request->getUri()->getPath() and $this->uri->getPath() to variables to not execute the code twice, in the expected case of having to add the prefix?

joelwurtz:
I agree with @dbu, removing the request identifier when a response is resolved should be a better way of doing this (if it's work of course).

magnusnordlander:
We're running in to an issue which we think might be caused by #103, and should probably be mitigated by this.

Really, the root of the problem is that `spl_object_hash` is not appropriate to use in situations like this (and there are no decent substitutes). `spl_object_hash` is guaranteed to be the same when called twice on an object, but it is not guaranteed to be unique, because hashes may be reused after objects have been deallocated.

From the manual:

> This function returns a unique identifier for the object. This id can be used as a hash key for storing objects, or for identifying an object, as long as the object is not destroyed. Once the object is destroyed, its hash may be reused for other objects.

One workaround would be to store a reference to the object, which of course would cause it not to be deallocated, but the downside is of course increased memory usage.

nicholasruunu:
One fix to the problem (^) might be to save the altered request uris instead and check:
`if (!array_key_exists($request->getUri()->getPath(), $this->alteredRequests)`

It could possibly create some kind of issue I haven't thought about.

dbu:
i like that header idea joel. does it mean the plugins potentially need to share that header? alternatively, we could define that the plugin client will scan the headers and remove any headers starting with `X-HTTPlug-PluginMetadata-` so that each plugin can have its own metadata headers...

nicholasruunu:
Sounds good @joelwurtz, but do we really need an id for this?
`X-HTTPlug-PluginMetadata: addPath-altered=1` in the request should be enough?
Then just remove X-HTTPlug-PluginMetadata all together.

If that's too much work to check there's I think @dbu's solution will suffice but with `X-HTTPlug-PluginMetadata-addPath-altered: 1`

joelwurtz:
yes it would be fine @nicholasruunu, certainly it does not need a header

@dbu i think we can start with a single header, as a header can have multiple value it should be fine ATM. We can always change that later (i would make this internal first, external plugin should not use it : we dont provide bc contract)

dbu:
>
>
> X-HTTPlug-PluginMetadata: addPath-altered=1
>
That would be a convention. Maybe that is good enough. Can plugins add such
values with psr-7 methods or does it need string parsing?

mxr576:
> do you want to start a new pull request for this?
@dbu I'd because my library installs this PR as a patch and I would like test the described solution first in IRL to be sure that it actually solves this problem.

magnusnordlander:
Using a single (non-unique) flag will mean that there can't be multiple AddPathPlugins. I'm not sure when you'd ever have that, but it is technically possible, so I'd argue that either the flag name should be unique (e.g. addPath-altered-34dfd3, where the last part has been generated randomly in the plugin constructor), or that it should have a value unique to the plugin instance.

dbu:
oh, good point. the spl_object_hash should be good for this one, right? plugins won't be added and removed from the chain on the fly, so no spl hash collision risk i think. we could use `addPath-altered=spl_object_hash($this)` maybe?

nicholasruunu:
@dbu I think that `uniqid()` will be fine here. There will be no need to compare it to anything but the value itself.

magnusnordlander:
`uniqid()` isn't necessarily a bad choice, but it sleeps for a microsecond when you call it though. That's probably not an issue here, but if it is, `mt_rand()` would be an alternative that doesn't sleep. `spl_object_hash($this)` is probably a no-go here too, since the value is stored in something that does not have an object reference back.

dbu:
then i vote for `mt_rand`

joelwurtz:
mt_rand is not unique, so you can have the same problem as without it.
We can simply use a private static member incremented by one each time we used it: fast, no sleep and unique.

nicholasruunu:
Yeah, I agree that `mt_rand` feels wrong however unlikely it is to fail.
Making a static incrementing method also feel ugly, but I don't have much better options with the current design.
So to solve this pressing issue I'd vote for this option.

I think for the original issue (#103) of being able to resend an altered request I think this is the wrong way to go though. What you want is to resend the original request, not let all the plugins figure out if it's unaltered or not.

dbu:
@nicholasruunu i went looking up the code and the reason this happens like this is that the plugin method is [passed $first](https://github.com/php-http/client-common/blob/e35c5f3f74707760e1ced52f8533121b8072feb5/src/Plugin.php#L31) which is the first plugin, for when they want to "restart" the request. the plugins only have the request at its current state, not the original unmodified request. so when they want to restart for whatever reason, all plugins are applied a second time to the request. so i think we do need the discussed solution.

@mxr576 do you have time for a pull request to implement the solution with the static increment?

dbu:
this is how the plugin chain is created: https://github.com/php-http/client-common/blob/e35c5f3f74707760e1ced52f8533121b8072feb5/src/PluginClient.php#L145

the plugin does not have access to the original request, and the "first" plugin is not hardwired with the unchanged request. therefore if a plugin restarts the chain, it only has access to the request as it currently looks. changing this would be quite a big change for how the plugin system works.

nicholasruunu:
@dbu Yeah I see what you mean... this package is breaking my brain. 😩
My thinking is that AddPath must be a special kind of plugin that only runs on requests sent from the user and only once. Or not being a plugin at all.

dbu:
another thought: adding a path only makes sense for known hosts. we could change the add path plugin to be configured with a regex for the domains for which to add a prefix.

the only place we ever call $first is in the RedirectPlugin. retry for example does not restart from the beginning, but from the current location. therefore i think instead of trying to track the chain, i'd make AddPath look at the domain and assume that no plugin restarts the same request with the same path. (if authentication needs a separate request, that should not be done with a plugin that restarts the request to do an authentication request, but by implementing Authentication and injecting a client to the auth implementation)

Nyholm:
I added an alternative solution here: #141

Nyholm:
There are no good solution. Just solutions that redefine what edge cases we will support. 

In 2.x will have a simple approach. See #141 

Im closing this as "wont fix"

PR #114, opened 2018-10-31
Allow HTTPlug 2.0

Jean85:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | not sure yet
| Deprecations?   | no
| Related tickets | -
| Documentation   | not sure it's needed yet
| License         | MIT

#### What's in this PR?
This PR (like https://github.com/php-http/discovery/pull/115) is to test and allow HTTPlug 2 (and PSR-18)

#### Checklist

#### To Do
- [ ] Check and fix build failures

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,7 @@ matrix:
 
     allow_failures:
         - php: 7.3
+          sudo: required
         - env: STABILITY="dev"
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ matrix:
         - php: 7.1
         - php: 7.2
         - php: 7.2
-          env: COVERAGE=true TEST_COMMAND="composer test-ci"  DEPENDENCIES="henrikbjorn/phpspec-code-coverage:^3.0"
+          env: COVERAGE=true TEST_COMMAND="composer test-ci"  DEPENDENCIES="leanphp/phpspec-code-coverage"
 
           # Test LTS versions
         - php: 7.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,8 @@ branches:
 matrix:
     fast_finish: true
     include:
-        - php: 7.1
-          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" DEPENDENCIES="doctrine/instantiator:^1.1"
+        - php: 7.0
+          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
 
           # Test the latest stable release
         - php: 7.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ matrix:
         - php: 7.0
         - php: 7.1
         - php: 7.2
+        - php: 7.2
           env: COVERAGE=true TEST_COMMAND="composer test-ci"  DEPENDENCIES="henrikbjorn/phpspec-code-coverage:^3.0"
 
           # Test LTS versions
@@ -38,9 +39,9 @@ matrix:
           env: STABILITY="dev"
 
     allow_failures:
-          # Latest dev is allowed to fail.
+        - php: 7.3
         - env: STABILITY="dev"
-          
+
 before_install:
     - if [[ $COVERAGE != true ]]; then phpenv config-rm xdebug.ini || true; fi
     - if ! [ -z "$STABILITY" ]; then composer config minimum-stability ${STABILITY}; fi;
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 language: php
-
 sudo: false
-
-dist: trusty
 
 cache:
     directories:
@@ -23,9 +20,6 @@ matrix:
           env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" DEPENDENCIES="doctrine/instantiator:^1.1"
 
           # Test the latest stable release
-        - php: 5.4
-        - php: 5.5
-        - php: 5.6
         - php: 7.0
         - php: 7.1
         - php: 7.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ matrix:
         - php: 7.0
         - php: 7.1
         - php: 7.2
-          env: COVERAGE=true TEST_COMMAND="composer test-ci"  DEPENDENCIES="henrikbjorn/phpspec-code-coverage:^1.0"
+          env: COVERAGE=true TEST_COMMAND="composer test-ci"  DEPENDENCIES="henrikbjorn/phpspec-code-coverage:^3.0"
 
           # Test LTS versions
         - php: 7.1
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     ],
     "require": {
         "php": "^5.4 || ^7.0",
-        "php-http/httplug": "^1.1",
+        "php-http/httplug": "^1.1|^2.0",
         "php-http/message-factory": "^1.0",
         "php-http/message": "^1.6",
         "symfony/options-resolver": "^2.6 || ^3.0 || ^4.0"
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,8 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0-dev"
+            "dev-2.x": "2.x-dev",
+            "dev-master": "1.8-dev"
         }
     }
 }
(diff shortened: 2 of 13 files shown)


composer.json
```diff
@@ -12,7 +12,7 @@
     ],
     "require": {
         "php": "^5.4 || ^7.0",
-        "php-http/httplug": "^1.1",
+        "php-http/httplug": "^1.1|^2.0",
```
Nyholm: FYI, The "official" syntax is `"^1.1 || ^2.0"`

dbu:
have a composer --prefer-lowest build with php 7.3? or do we need more than that?

Jean85:
Build is failing because `HttpMethodsClient::sendRequest` is missing the return type, so it's not PSR-18 and HTTPlug 2 compatible.

How should I tackle this? I see two roads:
* aim to a new major release compatible just with HTTPlug 2, and leave the resolution of conflicting versions to Composer
* try to work some magic and use some backward-compat trick to avoid any breaking changes

I would prefer the former, since I don't see how we could implement the latter; it would work only on PHP 7.2+ which has parameter widening, but we will be out of luck in PHP 7/7.1

dbu:
:+1: for a new major version that only works with httplug 2. will also keep the testing simpler ;-)

sagikazarmark:
New major version is OK, but that should go to a 2.x branch as we will probably have multiple, subsequent PRs.

Jean85:
@sagikazarmark perfectly ok with that! If you can create the branch I'll re-target this PR and add (instead of replace) the branch alias in `composer.json`

sagikazarmark:
2.x branch is ready ;)

Jean85:
Current build failure are pretty strange, because they manifest only when restricting the deps with external deps: https://travis-ci.org/php-http/client-common/builds/448853233

Same config without additional requires are green. I'll try to compare the installed deps to find the differences.

Jean85:
I have different results between CI and local. The base problem seems to be in `BatchExceptionSpec`, because it's trying to test (with mocking or similar stuff) a class (`BatchException`) which is both final and an exception, so not doable.

What should I do? I'm not well versed with PHPSpec...

.travis.yml
```diff
@@ -19,17 +16,15 @@ branches:
 matrix:
     fast_finish: true
     include:
-        - php: 7.1
-          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" DEPENDENCIES="doctrine/instantiator:^1.1"
+        - php: 7.0
```
Jean85: This was previously on PHP 7.1 due to doctrine/istantiator 1.1, which is no longer required. I've added a conflict rule below.

.travis.yml
```diff
@@ -19,17 +16,15 @@ branches:
...
           # Test the latest stable release
-        - php: 5.4
-        - php: 5.5
-        - php: 5.6
         - php: 7.0
         - php: 7.1
         - php: 7.2
-          env: COVERAGE=true TEST_COMMAND="composer test-ci"  DEPENDENCIES="henrikbjorn/phpspec-code-coverage:^1.0"
+        - php: 7.2
+          env: COVERAGE=true TEST_COMMAND="composer test-ci"  DEPENDENCIES="leanphp/phpspec-code-coverage"
```
Jean85: This is a maintained fork of the previous package, which is now abandoned.

.travis.yml
```diff
@@ -44,9 +39,10 @@ matrix:
           env: STABILITY="dev"
 
     allow_failures:
-          # Latest dev is allowed to fail.
+        - php: 7.3
```
Jean85: This is not working for now, but the issue is on Travis' side. Will work as soon as they fix it.

composer.json
```diff
@@ -11,16 +11,21 @@
...
-        "php-http/httplug": "^1.1",
+        "php": "^7.0",
+        "php-http/httplug": "^2.0",
         "php-http/message-factory": "^1.0",
         "php-http/message": "^1.6",
         "symfony/options-resolver": "^2.6 || ^3.0 || ^4.0"
     },
     "require-dev": {
-        "phpspec/phpspec": "^2.5 || ^3.4 || ^4.2",
+        "phpspec/phpspec": "^3.4 || ^4.2",
```
Jean85: Dropping 2.x which was needed only to run under PHP 5.x

composer.json
```diff
@@ -11,16 +11,21 @@
...
         "guzzlehttp/psr7": "^1.4"
     },
+    "conflict": {
```
Jean85: Those are lower bounds to avoid issues in CI builds. Should not affect `--prefer-lowest`.

composer.json
```diff
@@ -19,11 +19,11 @@
     },
     "require-dev": {
         "phpspec/phpspec": "^3.4 || ^4.2",
-        "phpspec/prophecy": "^1.8",
         "guzzlehttp/psr7": "^1.4"
     },
     "conflict": {
         "doctrine/instantiator": "<1.0.5",
+        "phpspec/prophecy": "<1.8",
```
joelwurtz: Does conflict interact with the global require ? Because it's only in conflict for the dev environment, does conflict-dev exist ?

joelwurtz: Sadly conflict-dev does not exist, so we should keep the old behavior (minimum version in require dev), as the runtime library does not conflict with prophecy < 1.8. It's only the tests.

GrahamCampbell: It's possible to simulate conflict-dev by making a dummy package that has exactly one conflict, then require said package in require-dev.

dbu: good catch. i'd prefer requiring prohecy ^1.8 in require-dev over creating a dummy package.

composer.json
```diff
@@ -11,16 +11,21 @@
         }
     ],
     "require": {
-        "php": "^5.4 || ^7.0",
-        "php-http/httplug": "^1.1",
+        "php": "^7.0",
```
dbu: why not go directly to 7.1? afaik 7.0 is EOL already.

composer.json
dbu: --prefer-lowest will respect the conflict section. as will other packages, as discussed below. do we use the doctrine/instantiator or sebastian/comparator in our actual code, or only for tests? all things thests should be solved in require-dev rather than here, or we prevent installing this on some systems that are fine with those versions.

Jean85: Both packages are needed by the test suite through PHPSpec/Prophecy; the dev deps tree should not have packages in common with the non-dev deps, so it should be fine. I preferred to use conflicts instead of requiring them explicitly because they are both indirect dependencies that could go away at any time.

composer.json
Jean85: Ouch I didn't think of this issue! Maybe moving to require-dev is good enough?

composer.json
Jean85: Because it's the same of PSR-18 and HTTPlug 2.

Nyholm: Im fine with 7.1. For psr18 it did not really matter. It was not using any php7.1 features.

sagikazarmark:
> We could modify it to allow httplug 1 and 2. That would be an easier migration, right?

Technically that could work. We should probably consider it.

joelwurtz:
Does not work since HttpMethodsClient is not final (so it's a BC break if we change the signature, exactly like the Guzzle package)

Nyholm:
You are correct. I did not consider that this library actually have http-client implementations.

Jean85:
Yep, exactly, it's a BC. But users of this lib will be able to require both versions if they want to support both HTTPlug versions.

BTW, what do you want me to do regarding the lower bound for Prophecy and doctrine/instantiator?

composer.json
dbu: as long as we agree that dropping 7.0 support is a minor change and we "just do it" if 7.0 becomes a pain for some reason, i am fine with allowing it for now ;-)

composer.json
dbu: the problem is that conflicts are also respected by applications using this package. meaning that if my project uses an older version of one of the libraries we mark as conflicting here will not be able to install this package because of the conflict.

conflict should only be used to mark versions of packages that may not be run together with this package. they should not be used to control what is installed for development. if this is all only about --prefer-lowest, i think we either need them in require-dev, or if you want to be more explicit, put them in the DEPENDENCIES of .travis.yml only for the lowest version build - that would be the most explicit way of doing it, imho.

composer.json
joelwurtz: Yes please set thoses packages into the require dev section instead of using conflict

Jean85:
I've moved the conflicts to require-dev as per https://github.com/php-http/client-common/pull/114#discussion_r230779636

Issue #115, opened 2018-11-06
Release 2.0


Jean85:
During #114 we started working toward a 2.0, with a dedicated branch. With that merged, 2.0 is compatible with HTTPlug 2 and PSR-18. What else do we want to add? I'll try to suggest some stuff and track it here. Maybe we can use a milestone?

- [ ] Update the docs/changelog
- [x] Use PHP 7 language features, since it's now the minimum
  - [x] Scalar type in method parameters and return types
- [x] Remove deprecations
- [x] Make all classes that are not intended to be extended final, verify that everything that is protected really needs to be protected

sagikazarmark:
Good idea, added something too.

I would probably wait with tagging until we actually have a client that we can use with this package, but Guzzle adapters will probably be tagged by then.

dbu:
:+1: i edited the PHP 7 thing to be more concrete. did you have other things in mind? we should focus on potentially BC breaking things. we can optimize implementation constructs after 2.0 is released, but method signatures, visibility/removing deprecations and final etc are BC breaks that we want to do now.

Jean85:
We should add #20 to this list.

dbu:
there is https://github.com/php-http/client-common/milestone/3 - imho we should use the milestone to group all tasks for the 2.0 release.

mxr576:
Any ETA for the 2.0.0 release?

Nyholm:
We will release it before Wednesday. 
@mxr576 Do you want to help review it and make sure every thing is correct?

mxr576:
Yes, I am happy to help. I am also updating my library's dependencies in this week.
I'll give it to try to #141 which was a follow up on #113. Of course, validate all other changes in the master branch with my libs automated tests.

Nyholm:
That would be great. 

Im pretty sure that current dev-master will be tagged 2.0. We just need more people testing it and review it.

dbu:
i tagged 2.0.0-RC1 just now. please report issues if you can find any.

mxr576:
Type-safe coding is an essential and a long awaited feature in PHP but it is ... when you trying to support multiple major versions from the same library. It seems my lib can not support both the 1.x and 2.x version from this library within the current 1.x major release thanks to parameter type changes like this: https://github.com/php-http/client-common/commit/68713627e85434c1a20529e8c108ce0c0938950b#diff-3f5c1f14c77b75e280cc2e297cce7d70R32.
[...]

Issue #20, opened 2016-03-31
Make clients final


sagikazarmark:
| Q | A |
| --- | --- |
| Bug? | no |
| New Feature? | yes |
| Version | 2.0.0 |

Make the rest of clients final. See #18 and #19

- [x] `BatchClient`
- [x] `BatchResult`
- [x] `Deferred`
- [x] `LeastUsedClientPool`
- [x] `RoundRobinClientPool`
- [x] `HttpMethodsClient`
- [x] `PluginClient`
- [x] `PluginClientFactory`
- [x] All Plugins

dbu:
updated description, i think we should also make all other classes that are not meant to be extended final.

Jean85:
I was trying to address this issue, but making the last non-final plugin final (RedirectPlugin) is pretty hard, since the class is extended in the specs to create stubs. This is obviously wrong because it tries to test internal behavior, but I'm not be able to solve that easily. Anyone else up for this?

dbu:
there is https://github.com/php-http/client-common/milestone/3 - imho we should use the milestone to group all tasks for the 2.0 release.

dbu:
i guess thats why its not yet final :-/
can you do a pull request with all that works, and a note that the RedirectPlugin is missing? maybe mark can see if he finds a better way for the testing of that plugin?

sagikazarmark:
Yup, will take a look at it: #21

dbu:
i think we now made final everything that can be final

PR #117, opened 2018-11-19
[2.0] Add types

Jean85:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | partially #115 
| License         | MIT

#### What's in this PR?
Another PR toward 2.0

#### Why?
Because PHP 7.0 is now the required minimum.

#### Checklist

- [ ] Documentation pull request created (not needed?)

#### To Do
- [ ] Should I add `: Promise` where possible or it may interfere on how the promises works?

### DIFF
--- a/src/BatchClient.php
+++ b/src/BatchClient.php
@@ -52,7 +52,7 @@ class BatchClient implements HttpClient
      *                        BatchResult with a map of request to result for success, request to
      *                        exception for failures
      */
-    public function sendRequests(array $requests)
+    public function sendRequests(array $requests): BatchResult
     {
         $batchResult = new BatchResult();
 
--- a/src/BatchResult.php
+++ b/src/BatchResult.php
@@ -34,7 +34,7 @@ final class BatchResult
      *
      * @return bool
      */
-    public function hasResponses()
+    public function hasResponses(): bool
     {
         return $this->responses->count() > 0;
     }
@@ -44,7 +44,7 @@ final class BatchResult
      *
      * @return ResponseInterface[]
      */
-    public function getResponses()
+    public function getResponses(): array
     {
         $responses = [];
 
@@ -62,7 +62,7 @@ final class BatchResult
      *
      * @return bool
      */
-    public function isSuccessful(RequestInterface $request)
+    public function isSuccessful(RequestInterface $request): bool
     {
         return $this->responses->contains($request);
     }
@@ -76,7 +76,7 @@ final class BatchResult
      *
      * @throws \UnexpectedValueException If request was not part of the batch or failed
      */
-    public function getResponseFor(RequestInterface $request)
+    public function getResponseFor(RequestInterface $request): ResponseInterface
     {
         try {
             return $this->responses[$request];
@@ -93,7 +93,7 @@ final class BatchResult
      *
      * @return BatchResult the new BatchResult with this request-response pair added to it
      */
-    public function addResponse(RequestInterface $request, ResponseInterface $response)
+    public function addResponse(RequestInterface $request, ResponseInterface $response): BatchResult
     {
         $new = clone $this;
         $new->responses->attach($request, $response);
@@ -106,7 +106,7 @@ final class BatchResult
      *
      * @return bool
      */
-    public function hasExceptions()
+    public function hasExceptions(): bool
     {
         return $this->exceptions->count() > 0;
     }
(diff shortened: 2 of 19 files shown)


sagikazarmark:
I think we can also drop docblocks where the type is obvious from the hint.

Jean85:
Maybe we can do it with PHPCSFixer and the `no_superfluous_phpdoc_tags` + `no_empty_phpdoc` rules?

sagikazarmark:
Sounds like a good idea.

Jean85:
@sagikazarmark PHPCSFixer is not installed, and we have StyleCI... How should I proceed? Do we want to maintain StileCI? Or should we use PHPCSFixer all the way and add a separate job/stage on Travis?

I should open a separate PR for sure...

For now I've fixed the CS of this PR alone, for StyleCI.

sagikazarmark:
We are working towards pretty ci which relies on phpcsfixer config, so you are free to ignore styleci and use PHPCS fixer only.

Jean85:
I've fixed the stub and a few other things in 2a80705. I also took the liberty of adding the specs to the dev autoloader (236a8ea) and migrated to `::class` (e87785c).

Jean85:
After #120 this has a lot of conflicts. Fixing now...

Jean85:
I've fixed all the conflicts. I wanted to add `: Promise` to the `Plugin` interface but a lot of tests will fail :cry:

dbu:
looks good to me. 

i was wondering if we need a changelog entry, but i think we should make all those clients final classes and then the type declarations are no relevant change anymore. is there any classes in this PR that we explicitly should *not* declare final?

Jean85:
I don't know, you or @sagikazarmark should decide on that. I'll try lo list here all the impacted classes (checks will be used to mark classes made final):

- [ ] `BatchClient`: got `sendRequest(...): ResponseInterface` from PSR-18 and `sendRequests(...): BatchResult`
- [ ] `BatchResult`
- [ ] `Deferred`
- [ ] `HttpClientPool`, but it's abstract
- [ ] `LeastUsedClientPool`, extends `HttpClientPool`
- [ ] `RoundRobinClientPool`, extends `HttpClientPool`
- [ ] `HttpMethodsClient`
- [ ] `PluginClient`
- [ ] `PluginClientFactory`

Also, nearly all plugins are impacted, but I want to give another go to adding `: Promise` to the `Plugin` interface.

Jean85:
I fixed nearly all specs, but I'm not able to fix it for the `CookiePlugin`. It seems to me that the callbacks used as mock are not working properly:

```php
public function it_does_not_load_cookie_if_expired(RequestInterface $request, UriInterface $uri, Promise $promise)
{
    $cookie = new Cookie('name', 'value', null, 'test.com', false, false, null, (new \DateTime())->modify('-1 day'));
    $this->cookieJar->addCookie($cookie);

    $request->withAddedHeader('Cookie', 'name=value')->shouldNotBeCalled();

    $this->handleRequest($request, function (RequestInterface $requestReceived) use ($request, $promise) {
        if (Argument::is($requestReceived)->scoreArgument($request->getWrappedObject())) {
            return $promise->getWrappedObject();
        }
    }, function () {});
}
```
The error is:

> ----  broken examples
> 
>         Http/Client/Common/Plugin/CookiePlugin
>   79  ! does not load cookie if expired
>         exception [err:TypeError("Return value of Http\Client\Common\Plugin\CookiePlugin::handleRequest() must implement interface Http\Promise\Promise, null returned")] has been thrown.
> 

Anyone can help me here?

dbu:
lets do the `final` in a separate pull request. i only brought it up here because if we make the classes final we don't need a changelog.

dbu:
see also #20 => can you please add a changelog for the abstract class?

@sagikazarmark is our expert in phpspec, maybe he can help with the error?

Jean85:
@dbu changelog updated. I'll wait for help on the `CookiePlugin` specs :+1:

sagikazarmark:
I feel like being summoned. 😄 

Lemme check...

sagikazarmark:
It's one of my biggest regret that I didn't pay more attention to tests in this repo. We abused spec testing quite a bit....

Anyway, you could try using this line in the breaking examples:

`$this->handleRequest($request, PluginStub::next(), function () {});`

We don't really care about the returned promise in those cases.

PR #120, opened 2018-11-20
[2.0] Add PHP-CS-Ffixer

Jean85:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #115
| License         | MIT

#### What's in this PR?
As suggested in https://github.com/php-http/client-common/pull/117#issuecomment-439939188, we can add PHP-CS-Fixer for better DX an stricter CS.

I'm also migrating to Travis CI jobs structure, to allow better flow in the CS (and CS/AST checks too if needed).

#### Example Usage

```
$ composer cs-check
> vendor/bin/php-cs-fixer fix --dry-run
Loaded config default from "/home/alai/workspace/OSS/client-common/.php_cs".
Using cache file ".php_cs.cache".
   1) src/Deferred.php

Checked all files in 0.029 seconds, 10.000 MB memory used
Script vendor/bin/php-cs-fixer fix --dry-run handling the cs-check event returned with error code 8
```

#### To Do

- [ ] Add Pretty CI


dbu:
and imho we should delete .styleci.yml and disable the hook after this is merged.

Jean85:
StyleCI config removed. Let me know if I have to do anything else, or if we should add other CS rules.

I don't think I can add PrettyCI though. Should I add the CS check in Travis in the meantime?

dbu:
no, if you have run the php-cs-fixer on the cli that is good enough. i will merge this and enable pretty-ci.

dbu:
> the settings are slightly different from php-http/boilerplate#27 - i want to be consistent on all projects. but i think the fix here is to update the rules in boilerplate.

https://github.com/php-http/boilerplate/pull/28

dbu:
disabled styleci and enabled prettyci

PR #123, opened 2018-12-02
mark classes as final and methods and properties as private

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | partially #20 
| Documentation   | https://github.com/php-http/documentation/pull/244
| License         | MIT

#### What's in this PR?

Mark the remaining things final.

#### Why?

See #20

#### Checklist

### DIFF
--- a/src/BatchClient.php
+++ b/src/BatchClient.php
@@ -8,14 +8,7 @@ use Http\Client\Common\Exception\BatchException;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 
-/**
- * BatchClient allow to sends multiple request and retrieve a Batch Result.
- *
- * This implementation simply loops over the requests and uses sendRequest with each of them.
- *
- * @author Joel Wurtz <jwurtz@jolicode.com>
- */
-class BatchClient implements HttpClient
+final class BatchClient implements BatchClientInterface
 {
     /**
      * @var HttpClient
@@ -27,28 +20,11 @@ class BatchClient implements HttpClient
         $this->client = $client;
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function sendRequest(RequestInterface $request): ResponseInterface
     {
         return $this->client->sendRequest($request);
     }
 
-    /**
-     * Send several requests.
-     *
-     * You may not assume that the requests are executed in a particular order. If the order matters
-     * for your application, use sendRequest sequentially.
-     *
-     * @param RequestInterface[] The requests to send
-     *
-     * @return BatchResult Containing one result per request
-     *
-     * @throws BatchException If one or more requests fails. The exception gives access to the
-     *                        BatchResult with a map of request to result for success, request to
-     *                        exception for failures
-     */
     public function sendRequests(array $requests): BatchResult
     {
         $batchResult = new BatchResult();
--- a/src/HttpClientPool/HttpClientPoolItem.php
+++ b/src/HttpClientPool/HttpClientPoolItem.php
@@ -1,7 +1,8 @@
 <?php
 
-namespace Http\Client\Common;
+namespace Http\Client\Common\HttpClientPool;
 
+use Http\Client\Common\FlexibleHttpClient;
 use Http\Client\HttpAsyncClient;
 use Http\Client\HttpClient;
 use Psr\Http\Message\RequestInterface;
@@ -11,8 +12,15 @@ use Psr\Http\Message\ResponseInterface;
 /**
  * A HttpClientPoolItem represent a HttpClient inside a Pool.
  *
- * It is disabled when a request failed and can be reenable after a certain number of seconds
- * It also keep tracks of the current number of request the client is currently sending (only usable for async method)
+ * It is disabled when a request failed and can be reenabled after a certain number of seconds.
+ * It also keep tracks of the current number of open requests the client is currently being sending
+ * (only usable for async method).
+ *
+ * This class is used internally in the client pools and is not supposed to be used anywhere else.
+ *
+ * @final
+ *
+ * @internal
  *
  * @author Joel Wurtz <joel.wurtz@gmail.com>
  */
@@ -29,7 +37,11 @@ class HttpClientPoolItem implements HttpClient, HttpAsyncClient
     private $disabledAt;
 
     /**
-     * @var int|null Number of seconds after this client is reenable, by default null: never reenable this client
+     * Number of seconds until this client is enabled again after an error.
+     *
+     * null: never reenable this client.
+     *
+     * @var int|null
      */
     private $reenableAfter;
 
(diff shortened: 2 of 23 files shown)


src/BatchClient.php
```diff
@@ -15,7 +15,7 @@
  *
  * @author Joel Wurtz <jwurtz@jolicode.com>
  */
-class BatchClient implements HttpClient
+final class BatchClient implements HttpClient
```
gmponos: Batch client does not have an abstraction layer for all of it's methods.

src/HttpClientPoolItem.php
```diff
@@ -16,7 +16,7 @@
  *
  * @author Joel Wurtz <joel.wurtz@gmail.com>
  */
-class HttpClientPoolItem implements HttpClient, HttpAsyncClient
+final class HttpClientPoolItem implements HttpClient, HttpAsyncClient
```
gmponos: Same.. this client also does not have an abstraction layer for all of it's methods.

gmponos:
I know that you want to enforce composition instead of inheritance but in real life scenarios there are cases that you will need to depend on implementation instead of abstraction.

This way it will make mocking impossible. Especially when clients like `BatchClient` do not have an abstraction layer for all of their methods.

Don't get me wrong I am all in for final but not for packages like this. If it was like guzzle having only one implementation I think it would make sense. But for packages like this one that contain multiple implementations of the same abstraction I think it's not the best thing to do...

For instance [this](https://github.com/doctrine/cache/tree/master/lib/Doctrine/Common/Cache) no finals. Although I can't be sure if this was made on purpose. The point is that there will be real life scenarios that I will need to depend specifically in `ArrayCache`.. Same goes for clients inside this package.

dbu:
hm, i see what you mean, and with the utility clients like batch or method client, you are certainly right. but i think we still want to prevent people from extending the specific clients. our options are:

* mark them as `@internal` and if people rely on being able to extend the class, tell them "told you"
* provide interfaces for those clients

i think i prefer the second option, as that is cleaner.

@php-http/owners wdyt? 

@sagikazarmark and if you agree to make the clients final, can you check if its sane what i did with php spec?

Nyholm:
I agree with @gmponos. If we should be really correct, we should use an interface for all classes we mark as final. 

Interfaces could be added after 2.0.0 though. But (of course) we cannot add final.

sagikazarmark:
TBH I don't really like the `Impl` suffix. Unfortunately, I don't really have an alternative. :\

src/BatchClient.php
joelwurtz: :+1: maybe we can add final with annotation like Symfony does

src/HttpClientPool/HttpClientPoolItem.php
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * A HttpClientPoolItem represent a HttpClient inside a Pool.
+ *
+ * It is disabled when a request failed and can be reenabled after a certain number of seconds.
+ * It also keep tracks of the current number of open requests the client is currently being sending
+ * (only usable for async method).
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+interface HttpClientPoolItem extends HttpClient, HttpAsyncClient
```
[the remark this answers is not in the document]
joelwurtz: Not a fan of that, creating an interface because of mocking makes me sad, having an interface add a new dependency and i would not have that here, HttpClientPoolItem is an internal thing, so we should just tag it `@internal` or `@final` IMO

src/HttpMethodsClient.php
```diff
@@ -24,7 +24,7 @@
  * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
  * @author David Buchmann <mail@davidbu.ch>
  */
-class HttpMethodsClient implements HttpClient
+final class HttpMethodsClient implements HttpClient
```
joelwurtz: Same problem as the BatchClient here, not possible to mock public methods of this class, we should use an annotation here instead.

dbu:
i added the interfaces. i opted to change the classes to `Impl` and keep the name for the interface. this way, everywhere the class was typehinted, this will now point to the interface. the only change needed is in the bootstrap where classes are implemented.

i think the interfaces are a clean thing here and we want them. if somebody has a better idea for the naming, i am happy to change it, but i don't know one. we could have a subnamespace for the implementations, but that does not seem better :-)

if the other @php-http/owners agree, i will do a pull request on the documentation as well.

src/HttpClientPool/HttpClientPoolItem.php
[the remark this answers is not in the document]
dbu: yeah, you are right with that one. removed the interface and documented that this is internal only.

src/BatchClient.php
dbu: added an interface

src/HttpClientPoolItem.php
dbu: this should not be used outside of the library itself. but using annotations now so that we can still test it.

gmponos:
> TBH I don't really like the Impl suffix. Unfortunately, I don't really have an alternative. :\

It's the exact opposite of having `Interface` suffix to interfaces.

My point of view is that I would prefer to be inconsistent with interfaces and some to have the `interface` suffix and some will not.. rather with the classes... some to have the `impl` and some not.

src/BatchClientImpl.php
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Http\Client\Common;
+
+use Http\Client\Exception;
+use Http\Client\HttpClient;
+use Http\Client\Common\Exception\BatchException;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+final class BatchClientImpl implements BatchClient
```
gmponos: Maybe `SimpleBatchClient` just to remove the `impl` :/

gmponos: Or `CommonBatchClient`.. same suggestions for `router` also.

dbu: i like the "simple" prefix

dbu:
for php-http we decided to avoid the interface suffix. 

we could add it for those cases - it would also make the upgrade path a bit different, because there would be no BC break at all. instead, the user should change all of their code to typehint interface instead, but things work fine without, unless they want to test with mocks.

unless i hear any other arguments, i will go with the Interface suffix for these and chose ease of upgrade over consistency.

src/BatchClientImpl.php
dbu: we now use the Interface suffix for those

dbu:
okay, i changed to the interface suffix and like the changes and its impact a lot more! want to merge @sagikazarmark ?

sagikazarmark:
Sorry, looks good to me. 👍

PR #128, opened 2018-12-11
AddPathPlugin > Trim ending slash on path

ruudk:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

#### What's in this PR?

Instead of throwing `new \LogicException('URI path cannot end with a slash.');` when the URI path ends with a slash, this PR will rtrim it automatically.

#### Why?

Today was a day that we merged a PR that had successful CI builds but broke during deployment. The reason? The parameters set on the server ended with a `/`.

I don't really understand the reason for throwing exceptions this hard, while they can be recovered. Every time somebody converts a Guzzle client to HTTPlug I have to warn them that we first need to change the URI's everywhere to remove the ending slashes. And sometimes that breaks for no reason.

#### Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changed
 
 - When multiple cookies exist, a single header with all cookies is sent as per RFC 6265 Section 5.4
+- AddPathPlugin will now trim of ending slashes in paths
 
 ## 1.8.1 - 2018-10-09
 
--- a/spec/Plugin/AddPathPluginSpec.php
+++ b/spec/Plugin/AddPathPluginSpec.php
@@ -46,12 +46,24 @@ class AddPathPluginSpec extends ObjectBehavior
         $this->handleRequest($request, function () {}, function () {});
     }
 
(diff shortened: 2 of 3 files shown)


ruudk:
Pretty CI is failing because of:
```
Fatal error: Uncaught Error: Class 'SLLH\StyleCIBridge\ConfigBridge' not found in /tmp/code/.php_cs:13
```

gmponos:
I think I have seen some discussions going around in symfony related to trailing slashes: https://github.com/symfony/symfony/issues?utf8=%E2%9C%93&q=trailing

not sure if this could help get the importance of this change or if the discussions are unrelated with the change here...

dbu:
thanks. if symfony routing decides to ignore trailing slashes, we can too :-)

i tagged 1.8.2 with this and also merged master to 2.x so we will have this in the new major version.

PR #131, opened 2018-12-26
Remove php-cs-fixer from dependencies

Nyholm:
IMHO PHP-cs-fixer is not a dependency. It is a tool like PHPStrom. 
One should not require tools in composer.json.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,8 +22,7 @@
         "guzzlehttp/psr7": "^1.4",
         "phpspec/phpspec": "^3.4 || ^4.2",
         "phpspec/prophecy": ">=1.8",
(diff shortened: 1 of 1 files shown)


dbu:
we went back and forth on this i think. the problem is that cs fixer as well as phpunit sometimes introduce changes that make tests fail when you don't use the right version of the tool. as we use pretty-ci which uses its own version anyways, i am ok with this, however.

Nyholm:
Yeah, I understand. 

Currently we have a problem that php-cs-fixer does not support php7.3 so we cannot support php-73. That is all kinds of wrong.

PR #132, opened 2018-12-28
Added a way to be forward compatible with version 2.0

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

To avoid updating all our plugins and the HTTPlug bundle, we can be forward compatible. 
https://3v4l.org/pn47M
https://3v4l.org/2kgtG

### DIFF
--- a/src/Plugin/VersionBridgePlugin.php
+++ b/src/Plugin/VersionBridgePlugin.php
@@ -13,7 +13,7 @@ use Psr\Http\Message\RequestInterface;
  */
 abstract class VersionBridgePlugin implements Plugin
 {
-    abstract function doHandleRequest(RequestInterface $request, callable $next, callable $first);
+    abstract protected function doHandleRequest(RequestInterface $request, callable $next, callable $first);
 
     /**
      * {@inheritdoc}
--- a/src/Plugin/VersionBridgePlugin.php
+++ b/src/Plugin/VersionBridgePlugin.php
@@ -15,9 +15,6 @@ abstract class VersionBridgePlugin implements Plugin
 {
     abstract protected function doHandleRequest(RequestInterface $request, callable $next, callable $first);
 
-    /**
-     * {@inheritdoc}
-     */
     public function handleRequest(RequestInterface $request, callable $next, callable $first)
     {
         return $this->doHandleRequest($request, $next, $first);
--- a/src/Plugin/VersionBridgePlugin.php
+++ b/src/Plugin/VersionBridgePlugin.php
@@ -2,7 +2,6 @@
 
 namespace Http\Client\Common\Plugin;
 
-use Http\Client\Common\Plugin;
 use Psr\Http\Message\RequestInterface;
 
 /**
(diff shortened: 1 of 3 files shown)


dbu:
hm. i see what you want to do. but is it worth it? are the separate plugins that complicated? i'd rather tag a last 1.x version of those and then move to 2.x. there should be no other things depending on those plugins, so the cost of having a major version seems minimal to me.

Nyholm:
HttplugBundle is also effected since it has a custom plugin and a custom client.

GrahamCampbell:
I think this small BC layer is probably a good idea. Much less work for upgrading, and much less messy for trying to deal with 1.x and 2.x existing at the same time. Otherwise, we risk php-http ironically causing dependency conflicts and mess, with some people needing 1.x and others 2.x, purely because of the interface incompatibility of these 2 classes.

Nyholm:
Traits are better. Sure!

dbu:
looks good to me. pretty-ci seems to be a bit flaky these days :-(

Nyholm:
No, it is our config file in master: https://github.com/php-http/client-common/blob/master/.php_cs

This is fixed in 2.x

src/Plugin/VersionBridgePlugin.php
```diff
@@ -0,0 +1,21 @@
...
+ * A plugin that helps you migrate from php-http/client-common 1.x to 2.x. This
+ * will also help you to support PHP5 at the same time you support 2.x.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+trait VersionBridgePlugin
+{
+    abstract protected function doHandleRequest(RequestInterface $request, callable $next, callable $first);
+
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
```
xabbuh: Isn't this missing the return type hint?

src/VersionBridgeClient.php
```diff
@@ -0,0 +1,21 @@
...
+ * A client that helps you migrate from php-http/httplug 1.x to 2.x. This
+ * will also help you to support PHP5 at the same time you support 2.x.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+trait VersionBridgeClient
+{
+    abstract protected function doSendRequest(RequestInterface $request);
+
+    public function sendRequest(RequestInterface $request)
```
xabbuh: same here

src/Plugin/VersionBridgePlugin.php
Nyholm: This is for master (ie 1.9.0). 

Here we should be compatible with httplug 1. Check #133 for httplug 2.0.

xabbuh: Ah, I didn't see that PR yet. 👍

PR #135, opened 2018-12-28
Support PSR-18

### LINKED ISSUE #122: [2.0] Inject PSR18 Client instead of Httplug when using Decoration
So people not using our adapter / interface can still use this library

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

We should show that we support PSR-18. I've replaced `Http\Client\HttpClient` with `Psr\Http\Client\ClientInterface` on all the places where we consume a client. 

This will fix #122

### DIFF
--- a/src/FlexibleHttpClient.php
+++ b/src/FlexibleHttpClient.php
@@ -23,7 +23,7 @@ final class FlexibleHttpClient implements HttpClient, HttpAsyncClient
     public function __construct($client)
     {
         if (!($client instanceof ClientInterface) && !($client instanceof HttpAsyncClient)) {
-            throw new \LogicException('Client must be an instance of Http\\Client\\HttpClient or Http\\Client\\HttpAsyncClient');
+            throw new \LogicException('Client must be an instance of Psr\\Http\\Client\\ClientInterface or Http\\Client\\HttpAsyncClient');
         }
 
         $this->httpClient = $client;
--- a/src/FlexibleHttpClient.php
+++ b/src/FlexibleHttpClient.php
@@ -4,6 +4,7 @@ namespace Http\Client\Common;
 
 use Http\Client\HttpAsyncClient;
 use Http\Client\HttpClient;
+use Psr\Http\Client\ClientInterface;
 
 /**
  * A flexible http client, which implements both interface and will emulate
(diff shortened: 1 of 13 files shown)


src/PluginClientFactory.php
```diff
@@ -33,9 +32,9 @@ public static function setFactory(callable $factory)
     }
 
     /**
-     * @param HttpClient|HttpAsyncClient $client
-     * @param Plugin[]                   $plugins
-     * @param array                      $options {
+     * @param ClientInterface|HttpAsyncClient $client
```
sagikazarmark: Missing import?

GrahamCampbell: https://github.com/GrahamCampbell/Analyzer would have detected that. It can be setup simply as a phpunit test: https://github.com/GrahamCampbell/Laravel-GitHub/blob/master/tests/AnalysisTest.php.

src/FlexibleHttpClient.php
```diff
@@ -17,18 +18,18 @@
...
 
     /**
-     * @param HttpClient|HttpAsyncClient $client
+     * @param ClientInterface|HttpAsyncClient $client
      */
     public function __construct($client)
     {
-        if (!($client instanceof HttpClient) && !($client instanceof HttpAsyncClient)) {
+        if (!($client instanceof ClientInterface) && !($client instanceof HttpAsyncClient)) {
             throw new \LogicException('Client must be an instance of Http\\Client\\HttpClient or Http\\Client\\HttpAsyncClient');
```
dbu: we should adjust the exception message too

src/Plugin/AddPathPlugin.php
```diff
@@ -47,7 +47,8 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
         $identifier = spl_object_hash((object) $first);
 
         if (!array_key_exists($identifier, $this->alteredRequests)) {
-            $request = $request->withUri($request->getUri()
+            $request = $request->withUri(
+                $request->getUri()
                 ->withPath($this->uri->getPath().$request->getUri()->getPath())
```
dbu: not sure that this is more readable than the previous formatting. should we indent this some more? or extract the argument on this line to a variable so we can put `$request->getUri()->withPath($prefixedPath)` on one line?

Nyholm:
Thank you for the review

src/PluginClientFactory.php
dbu: nice! 
phpstan would also detect this i think, and some other semantic mistakes. on a high enough strictness even things like type mixups as in https://github.com/php-http/discovery/pull/112

dbu: the import is fixed. if somebody wants to add the analyzer (or phpstan) lets do that in a separate pull request.

src/Plugin/AddPathPlugin.php
Nyholm: this change must have been suggested/automatic. 

I like your suggestion though.

Nyholm:
This will fix #122

PR #136, opened 2018-12-29
Fixed pretty-ci setup

Nyholm:


dbu:
can we disable the pretty-ci checks that force the additional cleanup? otherwise we should turn off pretty-ci in master completely to not have failing builds.

Nyholm:
Im not sure... 

In my mind we release 1.9, create a new branch from that and then merge 2.x. So Im fine with pretty CI is red for the last few PRs before 1.9

dbu:
if we can disable pretty-ci on the 1.9 or 1.x branch we create, i am fine with that. i want to avoid having red tests on the legacy version for as long as possible.

dbu:
this is obsolete as we move to 2.x now.

PR #137, opened 2018-12-29
Added declare strict types

gmponos:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

#### What's in this PR?

Added declare strict types on all files...

#### To Do

- Question: Should tests be included?

### DIFF
--- a/src/Plugin/CookiePlugin.php
+++ b/src/Plugin/CookiePlugin.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Http\Client\Common\Plugin;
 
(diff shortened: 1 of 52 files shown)


src/Plugin/CookiePlugin.php
```diff
@@ -127,7 +129,7 @@ private function createCookie(RequestInterface $request, $setCookie)
                                 $name,
                                 $value
                             ),
-                            null,
+                            0,
```
gmponos: Maybe this should be fixed on master first?

PR #139, opened 2018-12-29
retry http 5xx status as well

### LINKED ISSUE #126: Retry Plugin should also retry responses with status 5xx
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | 2

#### Actual Behavior

We only retry if there is an exception thrown.

#### Expected Behavior

We should also retry when a response arrives but the status is indicating a server error.

#### Possible Solutions

- a) Have an exception decider and a response decider (and same for delay)
- b) Expand the decider and delay functions to also pass them the response. This would be a hard BC break

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | fixes #126 
| Documentation   | https://github.com/php-http/documentation/pull/246
| License         | MIT

#### What's in this PR?

Add retry for responses with a server error status code.

#### Checklist

#### TODO

- [x] Rebase and cleanup when #142 has been merged to master and master to 2.x

### DIFF
--- a/src/Plugin/RetryPlugin.php
+++ b/src/Plugin/RetryPlugin.php
@@ -31,12 +31,22 @@ final class RetryPlugin implements Plugin
     /**
      * @var callable
      */
-    private $exceptionDelay;
+    private $errorResponseDelay;
+
+    /**
+     * @var callable
+     */
+    private $errorResponseDecider;
 
     /**
      * @var callable
      */
     private $exceptionDecider;
+
+    /**
+     * @var callable
+     */
+    private $exceptionDelay;
 
     /**
      * Store the retry counter for each request.
@@ -49,44 +59,39 @@ final class RetryPlugin implements Plugin
      * @param array $config {
      *
      *     @var int $retries Number of retries to attempt if an exception occurs before letting the exception bubble up
+     *     @var callable $error_response_decider A callback that gets a request and response to decide whether the request should be retried
      *     @var callable $exception_decider A callback that gets a request and an exception to decide after a failure whether the request should be retried
-     *     @var callable $exception_delay A callback that gets a request, an exception and the number of retries and returns how many microseconds we should wait before trying again
+     *     @var callable $error_response_delay A callback that gets a request and response and the current number of retries and returns how many microseconds we should wait before trying again
+     *     @var callable $exception_delay A callback that gets a request, an exception and the current number of retries and returns how many microseconds we should wait before trying again
      * }
      */
     public function __construct(array $config = [])
     {
-        if (array_key_exists('decider', $config)) {
-            if (array_key_exists('exception_decider', $config)) {
-                throw new \InvalidArgumentException('Do not set both the old "decider" and new "exception_decider" options');
-            }
-            trigger_error('The "decider" option has been deprecated in favour of "exception_decider"', E_USER_DEPRECATED);
-            $config['exception_decider'] = $config['decider'];
-            unset($config['decider']);
-        }
-        if (array_key_exists('delay', $config)) {
-            if (array_key_exists('exception_delay', $config)) {
-                throw new \InvalidArgumentException('Do not set both the old "delay" and new "exception_delay" options');
-            }
-            trigger_error('The "delay" option has been deprecated in favour of "exception_delay"', E_USER_DEPRECATED);
-            $config['exception_delay'] = $config['delay'];
-            unset($config['delay']);
-        }
-
         $resolver = new OptionsResolver();
         $resolver->setDefaults([
             'retries' => 1,
+            'error_response_decider' => function (RequestInterface $request, ResponseInterface $response) {
+                // do not retry client errors
+                return $response->getStatusCode() >= 500 && $response->getStatusCode() < 600;
+            },
             'exception_decider' => function (RequestInterface $request, Exception $e) {
                 // do not retry client errors
-                return !$e instanceof HttpException || $e->getCode() >= 500;
+                return !$e instanceof HttpException || $e->getCode() >= 500 && $e->getCode() < 600;
             },
-            'exception_delay' => __CLASS__.'::defaultDelay',
+            'error_response_delay' => __CLASS__.'::defaultErrorResponseDelay',
+            'exception_delay' => __CLASS__.'::defaultExceptionDelay',
         ]);
+
         $resolver->setAllowedTypes('retries', 'int');
+        $resolver->setAllowedTypes('error_response_decider', 'callable');
         $resolver->setAllowedTypes('exception_decider', 'callable');
+        $resolver->setAllowedTypes('error_response_delay', 'callable');
         $resolver->setAllowedTypes('exception_delay', 'callable');
         $options = $resolver->resolve($config);
 
         $this->retry = $options['retries'];
+        $this->errorResponseDecider = $options['error_response_decider'];
+        $this->errorResponseDelay = $options['error_response_delay'];
         $this->exceptionDecider = $options['exception_decider'];
         $this->exceptionDelay = $options['exception_delay'];
     }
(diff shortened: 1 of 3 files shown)


src/Plugin/RetryPlugin.php
```diff
@@ -47,29 +57,60 @@
...
-            'delay' => __CLASS__.'::defaultDelay',
+            'error_response_delay' => __CLASS__.'::defaultResponseDelay',
+            'exception_delay' => __CLASS__.'::defaultExceptionDelay',
         ]);
+
+        if (array_key_exists('decider', $config)) {
+            if (array_key_exists('exception_decider', $config)) {
+                throw new \InvalidArgumentException('Do not set both the old "decider" and new "exception_decider" options');
+            }
+            trigger_error('The "decider" option has been deprecated in favour of "exception_decider"', E_USER_DEPRECATED);
```
Nyholm: No need to have deprecated things here. 2.x is a fresh new branch

src/Plugin/RetryPlugin.php
```diff
@@ -47,29 +57,60 @@
...
+            $config['exception_decider'] = $config['decider'];
+            unset($config['decider']);
+        }
+        if (array_key_exists('delay', $config)) {
+            if (array_key_exists('exception_delay', $config)) {
+                throw new \InvalidArgumentException('Do not set both the old "delay" and new "exception_delay" options');
+            }
+            trigger_error('The "delay" option has been deprecated in favour of "exception_delay"', E_USER_DEPRECATED);
```
Nyholm: Same here

dbu:
i'll backport the rename to 1.9 with deprecation, and remove the BC code here

dbu:
rebased and cleaned up. this is now ready for review again.

src/Plugin/RetryPlugin.php
```diff
@@ -135,8 +152,29 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
     {
         return pow(2, $retries) * 500000;
     }
+
+    /**
+     * @param int $retries The number of retries we made before. First time this get called it will be 0.
+     *
+     * @return int
+     */
+    public static function defaultExceptionDelay(RequestInterface $request, Exception $e, $retries)
```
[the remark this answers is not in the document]
dbu: indeed. missed that when switching between php versions.

there are some other places that could have type declarations added. those have nothing to do with this change, so i won't do them here.

src/Plugin/RetryPlugin.php
```diff
@@ -120,23 +142,35 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
             }
 
             $time = call_user_func($this->exceptionDelay, $request, $exception, $this->retryStorage[$chainIdentifier]);
-            usleep($time);
-
-            // Retry synchronously
-            ++$this->retryStorage[$chainIdentifier];
-            $promise = $this->handleRequest($request, $next, $first);
 
-            return $promise->wait();
```
dbu: hm, did this ever work? this will return the actual response (unless there is an exception that is thrown, which also is wrong as it should be a rejected promise instead).

should we not call ourselves recursively instead?

dbu: ah wait. we are inside the `then` of the promise, so this is the correct thing to do, i think.

GrahamCampbell:
👍

src/Plugin/RetryPlugin.php
```diff
@@ -49,44 +59,41 @@
      * @param array $config {
      *
      *     @var int $retries Number of retries to attempt if an exception occurs before letting the exception bubble up
+     *     @var callable $error_response_decider A callback that gets a request and response to decide whether the request should be retried
      *     @var callable $exception_decider A callback that gets a request and an exception to decide after a failure whether the request should be retried
+     *     @var callable $error_response_delay A callback that gets a request and response and the number of retries and returns how many microseconds we should wait before trying again
      *     @var callable $exception_delay A callback that gets a request, an exception and the number of retries and returns how many microseconds we should wait before trying again
      * }
+     *
+     * @since exception_decider and error_response_decider have been added in version 2. In version 1.*, the exception_decider has been called decider.
```
Nyholm: Is this needed?

sagikazarmark:
Are you sure this is a good idea? 5xx generally means everything is broken and the server tells the client to leave them alone. Imagine if everyone retried requests upon an 500 error: poor server wouldn't have any chance to recover.

Nyholm:
What other request should you retry? 4xx means that the *request* itself is wrong.

sagikazarmark:
Not necessarily. 404 could mean a resource is not ready yet, but will be at some point.

I guess retry from application and retry from http client point of view mean two different things.

GrahamCampbell:
> Not necessarily. 404 could mean a resource is not ready yet, but will be at some point.

People can always configure reties themselves for that case though. For example, I do this when hitting the GitHub API, because they have a lag due to SQL replication.

dbu:
to me retrying 5xx seems reasonable, might be a temporary lapse in server communication or a database deadlock or whatnot. 

we can also by default not retry any response codes, if we think that is better. just having the configuration option has value already. what do we do?

src/Plugin/RetryPlugin.php
dbu: removed

Nyholm:
If you are using the RetryPlugin you are assuming that at least some request should be retried. 

I understand both arguments. I guess it depends on what kind of services you are using. I dont mind retrying my 500 requests when Im requesting github. But I may reconsider if I querying my own "easy to dos"-servers. 


I vote for a merge.

dbu:
> But I may reconsider if I querying my own "easy to dos"-servers.

we can add a warning in the documentation. there is the exponential backoff, so there is not that much requests to be expected.

dbu:
added documentation PR and realized i only want to retry codes from 500 to < 600. some systems (ab)use higher http status codes for custom things and we don't want to accidentally interact with those by default.

gmponos:
1. On this PR you are altering the default behavior of the plugin but a developer can set his own if he wishes to right? For what it is worth as a consumer developer I would never use this plugin without setting my custom callback functions... Therefore I would prefer if the default is `null`.
2. For the reason above I would prefer if the callback were allowed to be also `null` instead of having to pass a callback function with just a return...
3. For me this line `return !$e instanceof HttpException || $e->getCode() >= 500 && $e->getCode() < 600;` is confusing... It might make sense in the cases where you have a `PluginClient` with the `ErrorPlugin` inside of it... if the PluginClient becomes a PSR compatible client then according to PSR

> A Client MUST NOT treat a well-formed HTTP request or HTTP response as an error condition. For example, response status codes in the 400 and 500 range MUST NOT cause an exception and MUST be returned to the Calling Library as normal.

So I don't understand how this `$e->getCode() >= 500 && $e->getCode() < 600` can be fullfilled.. Because it is confusing I would really love if someone has the time to represent a full example with this. Maybe on a test...

On another topic please rethink the usage of `ErrorPlugin` with `PluginClient` on v2. On a discussion we had on chat if we initialize on v2 a PluginClient with the ErrorPlugin it wont be compatible with PSR-18

gmponos:
> For what it is worth as a consumer developer I would never use this plugin without setting my custom callback functions..

Elaborating this.. Using guzzle retry middleware... I only tend to retry 503 cases and also cases that the request was not even sent.. (status code 0 because of cURL)... Or cases that I got a response and I must look inside of the response...

dbu:
allowing to pass `null` makes the code more complicated. and the plugin is not active by default, so it can be used if wanted but is not needed. the exception part of the plugin indeed only makes sense with the error plugin - this is also explicitly mentioned in the [documentation](https://github.com/php-http/documentation/pull/246) to avoid confusion

following our slack discussion, i [added a note to the ErrorPlugin](https://github.com/php-http/client-common/pull/157) that indeed it is violating PSR-18. it can still be used when writing an application and wanting exceptions, but must not be used with third party libraries.

gmponos:
> allowing to pass null makes the code more complicated.

> and the plugin is not active by default

I understand these part..  None the less the `null` has to be taken in consideration along with the other parts that I said...

I saw also the documentation.. but I don't believe that the documentation here saves the day.. in overall, for what my opinion is worth it, I am not excited very much about these changes... :)

PR #141, opened 2018-12-29
Dont add path to an url we already added a path to.

### LINKED ISSUE #118: spl_object_hash collisions
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 1.8.0 and higher

#### Behavior
The method "spl_object_hash" can issue the same hash for two different objects **if** one of the objects has been destroyed (i.e. garbage collector run).

This hash collision can cause a request to pass without actually adding the path.

See the notes for "spl_object_hash" in the PHP documentation: http://php.net/manual/en/function.spl-object-hash.php 

#### Steps to Reproduce
Send multiple requests with the same HTTP client.

#### Possible Solutions
1. Do not cache requests (as done in 1.70)
2. Do not use 'spl_object_hash' for hashing objects

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #118, fixes #113
| Documentation   | 
| License         | MIT

#### What's in this PR?

What if we stored all the results of our rewrite and make sure we never rewrite a url that were stored as a result?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
   Added interfaces for BatchClient, HttpClientRouter and HttpMethodsClient.
   (These interfaces use the `Interface` suffix to avoid name collisions.)
 - Added an interface for HttpClientPool and moved the abstract class to the HttpClientPool sub namespace.
+- AddPathPlugin: Do not add the prefix if the URL already has the same prefix.
 
 ### Removed
 - Deprecated option `debug_plugins` has been removed from `PluginClient`
@@ -20,8 +21,6 @@
 
 - RetryPlugin: Renamed the configuration options for the exception retry callback from `decider` to `exception_decider`
   and `delay` to `exception_delay`. The old names still work but are deprecated.
-- AddPathPlugin: Do not add the prefix if the URL already has the same prefix.
-
 
 ## 1.8.2 - 2018-12-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,10 @@
 
 ### Changed
 
-- [RetryPlugin] Renamed the configuration options for the exception retry callback from `decider` to `exception_decider`
+- RetryPlugin: Renamed the configuration options for the exception retry callback from `decider` to `exception_decider`
   and `delay` to `exception_delay`. The old names still work but are deprecated.
+- AddPathPlugin: Do not add the prefix if the URL already has the same prefix.
+
 
 ## 1.8.2 - 2018-12-14
 
--- a/src/Plugin/AddPathPlugin.php
+++ b/src/Plugin/AddPathPlugin.php
@@ -21,9 +21,6 @@ final class AddPathPlugin implements Plugin
      */
     private $uri;
 
-    /**
-     * @param UriInterface $uri
-     */
     public function __construct(UriInterface $uri)
     {
         if ('' === $uri->getPath()) {
@@ -67,14 +64,14 @@ final class AddPathPlugin implements Plugin
      */
     public function handleRequest(RequestInterface $request, callable $next, callable $first): Promise
     {
-         $prepend = $this->uri->getPath();
-         $path = $request->getUri()->getPath();
+        $prepend = $this->uri->getPath();
+        $path = $request->getUri()->getPath();
 
-         if (substr($path, 0, strlen($prepend)) !== $prepend) {
-             $request = $request->withUri($request->getUri()
+        if (substr($path, 0, strlen($prepend)) !== $prepend) {
+            $request = $request->withUri($request->getUri()
                  ->withPath($prepend.$path)
              );
-         }
+        }
 
         return $next($request);
     }
--- a/src/Plugin/AddPathPlugin.php
+++ b/src/Plugin/AddPathPlugin.php
@@ -22,12 +22,8 @@ final class AddPathPlugin implements Plugin
     private $uri;
 
     /**
-     * Stores identifiers of the already altered requests.
-     *
-     * @var array
+     * @param UriInterface $uri
      */
-    private $alteredRequests = [];
-
     public function __construct(UriInterface $uri)
     {
         if ('' === $uri->getPath()) {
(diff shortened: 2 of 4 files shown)


mxr576:
I'll try out this patch next week. IIRC I tried a similar approach before without any luck when I worked on my library.

dbu:
that idea makes sense to me. we need to cleanup however, imagine i call the same url repeatedly in the same php process: the rewrite only happens once. with async, i am not sure we can even expect the promise to be resolved already. so we'd need to combine the spl hash and the url, and cleanup in the `then` of a promise. the request will live until the promise is resolved (because there is still a reference to it) so the spl hash can't be reused until then. tracking the url that we rewrote prevents us from rewriting it again.

Nyholm:
I dont understand, Why would it be any different for async?

Nyholm:
I still dont understand. Im sorry.

If i add the path `/api`

T0: An async request (A) to https://example.com/foo
T1: A normal request (B) to https://example.com/bar
T2: Request (B) will be rewritten to https://example.com/api/bar

T3: A normal request (C) to https://example.com/api/foo
T4: Request (C) will be rewritten to https://example.com/api/api/foo (this is expected)

T5: Request (A) will be resolved and rewritten to https://example.com/api/foo
T6: Same request (C) again to https://example.com/api/foo
T7: Request (C) will not be rewritten at all. It will go to https://example.com/api/foo

Issue. T4 and T7. 

-----------
Same would happen if: 

T0: A normal request (A) to https://example.com/api/foo
T1: Request (A) will be rewritten to https://example.com/api/api/foo 

T2: A normal request (B) to https://example.com/foo
T3: Request (B) will be rewritten to https://example.com/api/foo

T4: A normal request (C) to https://example.com/api/foo
T5: Request (C) will NOT be rewritten.
Issue: T1 and T5

---------

I mean, async or not does not matter.

Nyholm:
The issue above is acceptable. 
You use the `AddPath` plugin to always add a path. Ie, if your api endoints always starts with `/api/v1` then you may remove that from all your calls. 

My PR make sure we never add /api/v1` twice. 

The issue I describe is an edge case that will only happen if: 
- A) You call the API with your path included OR
- B) If you have two endpoints; one where the "path to add" is repeated and one where "path o add" is only once. Ie  
- - https://example.com/api/v1/api/v1/foo vs https://example.com/api/v1/foo  
- - https://example.com/api/api/foo vs https://example.com/api/foo

A is just using the plugin wrong. 
B is super unlikely. If you do have B you need to call them both and in *the wrong order* to have an issue.

dbu:
the current code will fail when we do

* Request http://example.com/path
* Request http://example.com/path

The first time, we add the prefix, the second time we don't. The plugin is not reset on each request.

If we add a `then` to remove the hash from our map when the request is finished, i think it will still not work if i do:

* Request A http://example.com/path
* Request B http://example.com/path
* Wait for A
* Wait for B

Nyholm:
> The first time, we add the prefix, the second time we don't. The plugin is not reset on each request.

No, I think you are reading the code wrong. 

I removed a test that was not working that probably confused you.

Nyholm:
I added a PHP unit test to prove that Im correct. 

My PHPspec knowledge is too poor. If anyone wants to write write my phpunit test to phpspec, then be my guest.

tests/Plugin/AddPathPluginTest.php
```diff
@@ -0,0 +1,136 @@
...
+    {
+        $verify = function (RequestInterface $request) {
+            $this->assertEquals('https://example.com/api/foo', $request->getUri()->__toString());
+        };
+
+        $request = new Request('GET', 'https://example.com/foo');
+        $this->plugin->handleRequest($request, $verify, $this->callable);
+
+        $request = new Request('GET', 'https://example.com/api/foo');
+        $this->plugin->handleRequest($request, $verify, $this->callable);
```
dbu: so in this case we do not add the prefix, even though we logically should. it seems wrong to me that the behaviour depends on the order of requests. if you reorder the 2, the test will fail because this thing gets rewritten to https://example.com/api/api/foo.

sagikazarmark:
I would also go with option a. It covers 98% of the use cases (`/api/api` is highly unlikely). There will always be cases that are too specific that we cannot support them.

Nyholm:
Sure! Let's do that!

src/Plugin/AddPathPlugin.php
```diff
@@ -46,14 +46,14 @@ public function __construct(UriInterface $uri)
      */
     public function handleRequest(RequestInterface $request, callable $next, callable $first)
     {
-        $identifier = spl_object_hash((object) $first);
-
-        if (!array_key_exists($identifier, $this->alteredRequests)) {
-            $request = $request->withUri($request->getUri()
-                ->withPath($this->uri->getPath().$request->getUri()->getPath())
-            );
-            $this->alteredRequests[$identifier] = $identifier;
```
dbu: please remove the unused field alteredRequests from the class

CHANGELOG.md
```diff
@@ -4,8 +4,10 @@
 
 ### Changed
 
-- [RetryPlugin] Renamed the configuration options for the exception retry callback from `decider` to `exception_decider`
+- RetryPlugin: Renamed the configuration options for the exception retry callback from `decider` to `exception_decider`
   and `delay` to `exception_delay`. The old names still work but are deprecated.
+- AddPathPlugin does not add prefix if the prefix already is defined.
```
dbu: maybe: "AddPathPlugin: Do not add the prefix if the URL already has the same prefix."

src/Plugin/AddPathPlugin.php
```diff
@@ -46,14 +46,14 @@ public function __construct(UriInterface $uri)
...
-        }
+         $prepend = $this->uri->getPath();
+         $path = $request->getUri()->getPath();
+
+         if (substr($path, 0, strlen($prepend)) !== $prepend) {
```
dbu: can we add a comment here with the gist of our reasoning why we ended up chosing this solution over more complicated things?

Nyholm:
How about this?

Nyholm:
> lets also update the documentation a bit to well explain the edge case where this now does not work as expected.

There is no documentation :/

src/Plugin/AddPathPlugin.php
```diff
@@ -42,18 +35,33 @@ public function __construct(UriInterface $uri)
...
+     * Some of the other edge cases are:
+     *  - request objects can be reused
+     *  - plugin chains could be restarted
+     *  - multiple AddPathPlugins may exist in one chain
+     *
+     * If one would like to cover all these scenarios one would have to add a custom
+     * header to the request to indicate it has been rewritten by the plugin. This
+     * implementation does not aim for pleasing all edge cases rather to be great
+     * value for the 95% other use cases.
+     *
```
dbu: thanks! i would prefer to be a bit more explicit:

```
    /**
     * Adds a prefix in the beginning of the URL's path.
     *
     * The prefix is not added if that prefix is already on the URL's path. This will fail on the edge
     * case of the prefix being repeated, for example if `https://example.com/api/api/foo` is a valid
     * URL on the server and the configured prefix is `/api`.
     *
     * We looked at other solutions, but they are all much more complicated, while still having edge
     * cases:
     *  - Doing an spl_object_hash on `$first` will lead to collisions over time because over time the
     *    hash can collide.
     * - Have the PluginClient provide a magic header to identify the request chain and only apply
     *   this plugin once.
     *
     * There are 2 reasons for the AddPathPlugin to be executed twice on the same request:
     * - A plugin can restart the chain by calling `$first`, e.g. redirect
     * - A plugin can call `$next` more than once, e.g. retry
     *
     * Depending on the scenario, the path should or should not be added. E.g. `$first` could
     * be called after a redirect response from the server. The server likely already has the
     * correct path.
     *
     * No solution fits all use cases. This implementation will work fine for the common use cases.
     * If you have a specific situation where this is not the right thing, you can build a custom plugin
     * that does exactly what you need.
     */

File: tests/Plugin/AddPathPluginTest.php [hunk 6]
```diff
@@ -0,0 +1,136 @@
...
+use Http\Client\Common\Plugin\AddPathPlugin;
+use Http\Client\Common\PluginClient;
+use Http\Client\HttpClient;
+use Nyholm\Psr7\Request;
+use Nyholm\Psr7\Response;
+use Nyholm\Psr7\Uri;
+use PHPUnit\Framework\TestCase;
+use Psr\Http\Message\RequestInterface;
+
+class AddPathPluginTest extends TestCase
```
dbu: i think we can simplify this test a lot now.

dbu:
> There is no documentation :/

found it here: https://github.com/php-http/documentation/blob/master/plugins/request-uri-manipulations.rst

Nyholm:
I rebased this PR on 2.x

dbu:
documentation: https://github.com/php-http/documentation/pull/245

nicholasruunu:
Good job guys. 👏

dbu:
thanks. this was really tricky - and i think we came full circle on the approach we use...

PR #142, opened 2018-12-30
rename decider to exception_decider to make room for response_decider in version 2

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | preparation for #139
| Documentation   | Callbacks are generally not documented at http://docs.php-http.org/en/latest/plugins/retry.html ...
| License         | MIT

#### What's in this PR?

Rename callback method in retry plugin options.

#### Why?

Make room for response decider in version 2.

#### Checklist

- not needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## 1.9.0 - [unreleased]
+
+### Changed
+
+- [RetryPlugin] Renamed the configuration options for the exception retry callback from `decider` to `exception_decider`
+  and `delay` to `exception_delay`. The old names still work but are deprecated.
+
 ## 1.8.2 - 2018-12-14
 
 ### Changed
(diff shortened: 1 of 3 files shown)


Nyholm:
Feel free to merge and rebase #139 on this.

mxr576:
Sorry to say, but the deprecation handling introduced in this PR is an overkill. As I can see it makes it impossible to support >= 1.9.0 and the older versions of this library by passing both the old and the new array keys to the plugin's construction. Why throwing an exception was necessary?
https://github.com/php-http/client-common/blob/1.9.0/src/Plugin/RetryPlugin.php#L55

dbu:
the exception is necessary because otherwise we would need to define a priority for redundant (and therefor potentially contradictory) options. and version 2 will no longer accept the old argument, while version 1.8 and older does not know anything about the new parameter.

i think for this, you will need to distinguish versions if you need to support both 1.9 and 2, if your code instantiates this and otherwise explain the configuration to the user.

PR #143, opened 2018-12-30
Make sure we can use PSR-18 clients in version 1.x

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

Our 2.x branch is compatible with PSR-18. The master branch should also be compatible. 

Ive ipdated all places where we are **consuming** a httplug client. They do now support both httplug 1.x and PSR-18.

FYI @gmponos

### DIFF
--- a/src/FlexibleHttpClient.php
+++ b/src/FlexibleHttpClient.php
@@ -18,7 +18,7 @@ final class FlexibleHttpClient implements HttpClient, HttpAsyncClient
     use HttpAsyncClientDecorator;
 
     /**
-     * @param HttpClient|HttpAsyncClient $client
+     * @param HttpClient|HttpAsyncClient|ClientInterface $client
      */
     public function __construct($client)
     {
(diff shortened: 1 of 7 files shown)


src/FlexibleHttpClient.php
```diff
@@ -21,14 +22,14 @@
      */
```
GrahamCampbell: param doc needs updating here i think?

gmponos:
Just throwing a long shot idea here.. on v1 what if there was a PSR18 HTTPlug adapter?

I think someone else has also suggested this but I am lost with all these issues..

Nyholm:
That would not help. Because PSR-18 has the exact same signature but with PHP7 return values. 

That is why all these updates to 2.x are super simple. They are just dropping PHP5 and adding return values all over the place.

src/FlexibleHttpClient.php
Nyholm: Thank you

Nyholm:
Thank you for merging

PR #144, opened 2018-12-31
Added test to highlight a bug

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

I found a bug. The bug is present all these are true: 
- We use a synchronous client
- We use a synchronous request
- We have a plugin that use `return $first($request)->wait()`

HOWEVER, using `return $first($request)->wait()` does not make any sense...
I got the idea from #103 but after writing this PR and https://github.com/php-http/client-common/pull/103#issuecomment-450636634 I discovered that there is no bug and one should never do `return $first($request)->wait()`. 

Im not sure we want to merge this test or not. I kind of like the code so I did not want to throw it away =)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,8 @@ jobs:
     env: DEPENDENCIES="dunglas/symfony-lock:^3"
   - php: 7.2
     env: DEPENDENCIES="dunglas/symfony-lock:^4"
+  - php: 7.2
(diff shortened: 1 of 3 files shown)


sagikazarmark:
To be honest, I think we should stop using phpspec. PHPSpec is great, but it requires too much hacks and results in absolutely not expressive tests in our case.

PR #145, opened 2019-01-02
add missing type declarations

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

#### What's in this PR?

Went over the code again to find places where we can add type hints.

### DIFF
--- a/src/Plugin/ContentTypePlugin.php
+++ b/src/Plugin/ContentTypePlugin.php
@@ -94,11 +94,11 @@ final class ContentTypePlugin implements Plugin
         return $next($request);
     }
 
-    /**
-     * @param $stream StreamInterface
-     */
-    private function isJson($stream): bool
+    private function isJson(StreamInterface $stream): bool
     {
+        if (!function_exists('json_decode')) {
+            return false;
+        }
         $stream->rewind();
 
         json_decode($stream->getContents());
@@ -108,17 +108,18 @@ final class ContentTypePlugin implements Plugin
 
     /**
      * @param $stream StreamInterface
-     *
-     * @return \SimpleXMLElement|false
      */
-    private function isXml($stream)
+    private function isXml(StreamInterface $stream): bool
     {
+        if (!function_exists('simplexml_load_string')) {
+            return false;
+        }
         $stream->rewind();
 
         $previousValue = libxml_use_internal_errors(true);
         $isXml = simplexml_load_string($stream->getContents());
         libxml_use_internal_errors($previousValue);
 
-        return $isXml;
+        return false !== $isXml;
     }
 }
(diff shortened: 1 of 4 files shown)


src/Plugin/ContentTypePlugin.php
```diff
@@ -94,11 +94,11 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
         return $next($request);
     }
 
-    /**
-     * @param $stream StreamInterface
-     */
-    private function isJson($stream): bool
+    private function isJson(StreamInterface $stream): bool
     {
+        if (!function_exists('json_decode')) {
```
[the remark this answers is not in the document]
GrahamCampbell: We should probably add ext-json to the suggest list of the composer.json, and describe where it's used, so people can decide if they need to install it.

src/Plugin/CookiePlugin.php
```diff
@@ -91,15 +91,13 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
      * Creates a cookie from a string.
      *
-     * @param $setCookie
-     *
      * @return Cookie|null
      *
      * @throws TransferException
      */
-    private function createCookie(RequestInterface $request, $setCookie)
+    private function createCookie(RequestInterface $request, string $setCookieHeader)
```
GrahamCampbell: `: ?Cookie`?

GrahamCampbell: I guess we'd need PHP 7.1 for that. Is there any reason why 2.x is php 7.0+ and not 7.1+. 7.0 is EOL already.

dbu: thats why i did not do it. but i wondered the same. happy to move to 7.1+. wdyt @Nyholm ?

Nyholm: I agree. Lets do PHP7.1

src/Plugin/ContentTypePlugin.php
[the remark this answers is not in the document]
dbu: good point. added ext-json and ext-libxml to suggest, with an explanation of what it would be used for.

src/Plugin/CookiePlugin.php
dbu: cool! however, i'd prefer to do that in a separate pull request, as it will entail quite a lot of changes.

dbu:
okay if i merge this and continue in a new PR to upgrade to php 7.1?

Nyholm:
Yes, go ahead.

Nyholm:
I added #148 about dropping PHP 7.0

PR #150, opened 2019-01-03
A better branch alias would be 2.0.x

GrahamCampbell:
Better, so that composer can't load the master branch in place of 2.1.x, since master is actually 2.0.x.


dbu:
thanks you!

PR #152, opened 2019-01-04
Add else option to the RequestMatcherPlugin

### LINKED ISSUE #65: Add else option to the RequestMatcherPlugin
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | 1.4

#### Actual Behavior

The request matcher checks if the given request matches the requirements, and executes the delegated plugin. In cases where we want to use an if-else logic to decide between two plugins this does not work well, and currently requires two RequestMatcher plugins.

#### Possible Solutions

Provide a second plugin to the RequestMatcherPlugin which gets called when the request is not a match.

It would also probably make sense to rename the RequestMatcher plugin to Conditional (this would be aligned with the Conditional authenticator in the message repo)

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #65
| Documentation   | 
| License         | MIT


Nyholm:
This will fail on PHP70.

Nyholm:
I've rebased. Waiting on travis

Nyholm:
Thank you for merging

PR #154, opened 2019-01-04
Only show Tavis build status for master

Nyholm:


Nyholm:
Thank you for the review.

PR #156, opened 2019-01-04
type declarations for php 7.1

dbu:
some more strict types where we can. the `: void` is useful for tools that report when the return value of a void method is being used.


Nyholm:
This needs to be cased to an int. 

https://github.com/php-http/client-common/blob/master/src/Plugin/RedirectPlugin.php#L142

dbu:
the problem are us mocking the response in a way that violates response interface. the status code must be an int according to the interface.

GrahamCampbell:
Looks pretty good. 👍

Nyholm:
Thank you

PR #157, opened 2019-01-06
add warning in error plugin

dbu:
warn that the error plugin violates PSR-18


gmponos:
What is your opinion about introducing an ErrorClient instead of an ErrorPlugin?

gmponos:
Sorry forget about doesn't make any difference.. Since HttpClient extends ClientInterface... so it must be PSR-18 either way...

dbu:
yeah, its a violation of the interface however we produce the exception. but in a smallish application that does not have proper domain exceptions, it can be "good enough" and convenient. if you do several web requests and want to abort the whole thing on (server) error, its easier than checking the return status after each request.

PR #158, opened 2019-01-06
Add plugin client builder

### LINKED ISSUE #96: Add a new plugin to the PluginClient
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | 1.7.0

#### Actual Behavior
Once your `PluginClient` has been initialized, you can't add new plugins to it

#### Expected Behavior
Being able to add some new plugins on the fly. I'm currently registering it in a Behat container, so I registered a few plugins on the construct https://github.com/Taluu/Behapi/blob/master/src/Container.php#L85.

But I would like to be able to give the possibility for the users to add some new plugins (such as the auth plugin or a random listener) so that they can do stuff on their own.

#### Possible Solutions
Add a `addPlugin` or something to the `PluginClient`

Taluu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #96 
| Documentation   | None, for now ?
| License         | MIT

#### What's in this PR?
This adds a plugin client builder, that allows to dynamically build a plugin client with plugins set on the go.

#### Why?
If you have some conditions on your plugin client (such as the auth can vary, some more headers can be dynamically added, ... like in the github's api from knp or my behat extension for http calls), it can be useful to not have a static plugin client with its plugin list, but to be able to add / remove them when needed.

As mentionned in the two repositories ([https://github.com/KnpLabs/php-github-api/blob/2.10.1/lib/Github/Client.php#L314-L331](Github's Client for the github php api of knp, or [my behat extension](https://github.com/Taluu/Behapi.git)), we both had to implement our own plugin builder, so this shows there's a "demand" for that.

We could have used the PluginClientFactory, but the problem with it is that when creating the client, we need to provide _all_ the plugins on the go, which is not (IMO) practical.

#### Example Usage

``` php
use Http\Client\Common\PluginClientBuilder;

// $client is a php-http client or a psr-18 one
// $plugin is a php-http plugin, pick one (auth basic, auth oauth, ... anything goes)

$builder = new PluginClientBuilder();
$builder->addPlugin($plugin); // by default, priority is 0

// do stuff....

$pluginClient = $builder->createClient($client);
// plugin client is a PluginClient with the `$plugin` plugin

// ...

$pluginClient = $otherBuilder->createClient($client);
// plugin client is another PluginClient and not the same as the one built previously
```

#### Checklist

#### To Do
- [x] Discuss the mutability ? **decision** : remove it.
- [x] How should we handle options in the builder ? In the create client ? In the constructor ? If in the create client, if changed should it retrigger the client creation ? **decision** : Add setter and remover for options, and inject that into the plugin client.
- [x] Add priorities in the plugins
- [x] Remove naming plugins
- [x] Tests ? As I didn't see many of those, and I wanted to throw a poc for a builder first

### DIFF
--- a/src/PluginClientBuilder.php
+++ b/src/PluginClientBuilder.php
@@ -1,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Http\Client\Common;
+
+use Http\Client\HttpAsyncClient;
+use Psr\Http\Client\ClientInterface;
+
+/**
+ * Build an instance of a PluginClient with a dynamic list of plugins.
+ *
+ * @author Baptiste Clavié <clavie.b@gmail.com>
+ */
+final class PluginClientBuilder
+{
+    /** @var Plugin[][] List of plugins ordered by priority [priority => Plugin[]]). */
+    private $plugins = [];
+
+    /** @var array Array of options to give to the plugin client */
+    private $options = [];
+
+    /**
+     * @param int $priority Priority of the plugin. The higher comes first.
+     */
+    public function addPlugin(Plugin $plugin, int $priority = 0): self
+    {
+        $this->plugins[$priority][] = $plugin;
+
+        return $this;
+    }
+
+    public function setOption(string $name, $value): self
+    {
+        $this->options[$name] = $value;
+
+        return $this;
+    }
+
+    public function removeOption(string $name): self
+    {
+        unset($this->options[$name]);
+
+        return $this;
+    }
+
+    /**
+     * @param ClientInterface | HttpAsyncClient $client
+     */
+    public function createClient($client): PluginClient
+    {
+        if (!$client instanceof ClientInterface && !$client instanceof HttpAsyncClient) {
+            throw new \RuntimeException('You must provide a valid http client');
+        }
+
+        $plugins = $this->plugins;
+
+        if (0 === count($plugins)) {
+            $plugins[] = [];
+        }
+
+        krsort($plugins);
+        $plugins = array_merge(...$plugins);
+
+        return new PluginClient(
+            $client,
+            array_values($plugins),
+            $this->options
+        );
+    }
+}
--- a/tests/PluginClientBuilderTest.php
+++ b/tests/PluginClientBuilderTest.php
@@ -49,6 +49,29 @@ class PluginClientBuilderTest extends TestCase
         $this->assertSame($expected, $plugged);
     }
 
+    /** @dataProvider clientProvider */
+    public function testOptions(string $client): void
+    {
+        $builder = new PluginClientBuilder();
+        $builder->setOption('max_restarts', 5);
+
+        $client = $this->prophesize($client)->reveal();
+        $client = $builder->createClient($client);
+
+        $closure = Closure::bind(
+            function (): array {
+                return $this->options;
+            },
+            $client,
+            PluginClient::class
+        );
+
+        $options = $closure();
+
+        $this->assertArrayHasKey('max_restarts', $options);
+        $this->assertSame(5, $options['max_restarts']);
+    }
+
     public function clientProvider(): iterable
     {
         yield 'sync\'d http client' => [HttpClient::class];
--- a/tests/PluginClientBuilderTest.php
+++ b/tests/PluginClientBuilderTest.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace tests\Http\Client\Common;
+
+use Closure;
+use Http\Client\Common\Plugin;
+use Http\Client\Common\PluginClient;
+use Http\Client\Common\PluginClientBuilder;
+use Http\Client\HttpAsyncClient;
+use Http\Client\HttpClient;
+use PHPUnit\Framework\TestCase;
+
+class PluginClientBuilderTest extends TestCase
+{
+    /** @dataProvider clientProvider */
+    public function testPriority(string $client): void
+    {
+        $builder = new PluginClientBuilder();
+
+        $plugins = [
+            10 => $this->prophesize(Plugin::class)->reveal(),
+            -10 => $this->prophesize(Plugin::class)->reveal(),
+            0 => $this->prophesize(Plugin::class)->reveal(),
+        ];
+
+        foreach ($plugins as $priority => $plugin) {
+            $builder->addPlugin($plugin, $priority);
+        }
+
+        $client = $this->prophesize($client)->reveal();
+        $client = $builder->createClient($client);
+
+        $closure = Closure::bind(
+            function (): array {
+                return $this->plugins;
+            },
+            $client,
+            PluginClient::class
+        );
+
+        $plugged = $closure();
+
+        $expected = $plugins;
+        krsort($expected);
+        $expected = array_values($expected);
+
+        $this->assertSame($expected, $plugged);
+    }
+
+    public function clientProvider(): iterable
+    {
+        yield 'sync\'d http client' => [HttpClient::class];
+        yield 'async\'d http client' => [HttpAsyncClient::class];
+    }
+}


src/PluginClientBuilder.php
```diff
@@ -0,0 +1,69 @@
...
+    /** @var ?PluginClient */
+    private $client;
+
+    public function addPlugin(Plugin $plugin): self
+    {
+        $builder = new self();
+
+        $builder->client = null;
+        $builder->plugins = $this->plugins;
+        $builder->plugins[get_class($plugin)] = $plugin;
```
sagikazarmark: This limits the user to add only one instance of the plugin to the chain. Even worse, there is no warning that a plugin is overwritten.

src/PluginClientBuilder.php
```diff
@@ -0,0 +1,69 @@
...
+{
+    /** @var Plugin[] */
+    private $plugins = [];
+
+    /** @var ?PluginClient */
+    private $client;
+
+    public function addPlugin(Plugin $plugin): self
+    {
+        $builder = new self();
```
sagikazarmark: Wonder why the builder is "immutable" here and it's not later.

src/PluginClientBuilder.php
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * @param ClientInterface | HttpClient | HttpAsyncClient $client Client to nest into the plugin client
+     */
+    public function createClient($client, array $options = []): PluginClient
+    {
+        if (!$client instanceof ClientInterface && !$client instanceof HttpClient && !$client instanceof HttpAsyncClient) {
+            throw new \RuntimeException('You must provide a valid http client');
+        }
+
+        if (null === $this->client) {
```
sagikazarmark: The builder is immutable at least it tries to be when it comes to modifying the plugin chain, but it isn't when you create a client. Also, if I want to use the builder multiple times and I pass in different client instances I won't get a plugin client with that specific client, but one with an eariler client instance without a warning. This kind of behavior easily leads to confusion IMHO.

src/PluginClientBuilder.php
Taluu: Originally I had a `string $name` argument, but in @Nyholm's pr on the builder, I saw he was using the fqcn instead.

I'll revert back to the plugin name instead.

src/PluginClientBuilder.php
Taluu: This is to avoid rebuilding the client each time. To solve this, I could add a method to set the http client on the builder itself (while making it immutable), and use that to build the plugin client (with a cache so it doesn't get build on each calls)

Taluu:
Updated with remarks from @sagikazarmark. I added the client in the constructor rather than the createClient to have a better "immutability" handling.

But with that, the `options` given in the `createClient` may pose problem if it changes... So we could either pass in into the constructor too, or find some sort of invalidating the built client.

Or we could add a setter for that one, as having options is, as its name implies it, optionnal.

src/PluginClientBuilder.php
```diff
@@ -0,0 +1,77 @@
...
+
+    /** @var ?PluginClient */
+    private $pluginClient;
+
+    /**
+     * @param ClientInterface | HttpClient | HttpAsyncClient $client Client to nest into the plugin client
+     */
+    public function __construct($client)
+    {
+        if (!$client instanceof ClientInterface && !$client instanceof HttpClient && !$client instanceof HttpAsyncClient) {
```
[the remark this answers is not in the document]
Taluu: If I understood @Nyholm, this is not specially for 2.0 (at least not blocking for it). I didn't check if there was a 1.x branch.

I'm not sure either if we should check for the `ClientInterface` though...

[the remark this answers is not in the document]
gmponos: - Master branch is v2
- Yes there is a branch for v1...
- Master branch which is v2 has as min requirement httplug: ^2.0 . The client on this version extends `ClientInterface`

> I'm not sure either if we should check for the ClientInterface though...

- Therefore my suggestion was to check only ClientInterface...

:)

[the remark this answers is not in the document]
gmponos: Checking all three would make more sense for v1.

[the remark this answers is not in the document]
Taluu: I'll retarget if there's hope for a 1.x release then. @Nyholm @sagikazarmark ?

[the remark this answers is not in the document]
gmponos: > Therefore my suggestion was to check only ClientInterface...

ofc you must keep the check for the Async client..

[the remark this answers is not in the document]
Taluu: why aren't you sure about that ?

sagikazarmark:
@Taluu I'm not against supporting this in 1.x as well per se, but 2.x is the new target of every feature now.

src/PluginClientBuilder.php
```diff
@@ -0,0 +1,77 @@
...
+final class PluginClientBuilder
+{
+    /** @var Plugin[] */
+    private $plugins = [];
+
+    /** @var ClientInterface | HttpClient | HttpAsyncClient */
+    private $client;
+
+    /** @var ?PluginClient */
+    private $pluginClient;
```
dbu: i agree with mark, would not cache the resulting client. we can create multiple instances of the same client with the builder, or even add more plugins and create another client with the additional plugins.

src/PluginClientBuilder.php
```diff
@@ -0,0 +1,77 @@
...
+        if (!$client instanceof ClientInterface && !$client instanceof HttpClient && !$client instanceof HttpAsyncClient) {
+            throw new \RuntimeException('You must provide a valid http client');
+        }
+
+        $this->client = $client;
+    }
+
+    public function addPlugin(string $name, Plugin $plugin): self
+    {
+        $builder = new self($this->client);
```
dbu: agree with mark here too - the builder should not be immutable but stateful. i found no good reference to support that claim, but all builders i have seen act that way. its also perfectly fine to have application code like

```
$builder = new PluginClientBuilder($client);
$builder->addPlugin('foo', new Plugin...);
```

src/PluginClientBuilder.php
```diff
@@ -0,0 +1,77 @@
...
+
+        $builder->pluginClient = null;
+        $builder->plugins = $this->plugins;
+        unset($builder->plugins[$name]);
+
+        return $builder;
+    }
+
+    public function createClient(array $options = []): PluginClient
```
dbu: i would prefer a `setOption(string $name, $value)` (and `removeOption(string $name)`) method instead. we right now don't have many options in the PluginClient, but it would be consistent imho.

Taluu:
I'll remove the immutability and the cache then. But for the plugin removal, it can be useful when we want to e.g remove a previous auth mechanism and replace it with another one. Unless the fact that a plugin in the bottom will somehow overwrite the previous one, I didn't check that case.

Adding `appendPlugin` and `prependPlugin` can be useful, I agree, but for `insertAfter` and `insertBefore`, I have some doubts, because as you said we would need to somehow know the plugins within the builder...

I'd rather add priorities actually. I'll try to work on that.

sagikazarmark:
Generally a builder serves one purpose. If you want for example create a base builder and use it later as the base for several other builders, you could add a `cloneBuilder` or `new` method that returns a new builder which can be modified without modifying a parent.

Priorities sound like and interesting feature. It can especially be useful when the building does not happen at one place (eg. the aformentioned DIC example). IIRC symfony event dispatcher has some sort of priority support.

Someone working on the HTTPlugBundle could jump in here as I imagine the bundle an similar solutions are first class citizens of this builder.

Taluu:
PR updated. I removed the immutability, and added priorities for plugins instead of names. I had to remove the possibility to remove a plugin, as we now have no means to remove a specific plugin (as there's no name given).

I chose to keep the fluent interface in the builder, as this is one of the few cases where it is legit. I'll update the first message with the examples.

src/PluginClientBuilder.php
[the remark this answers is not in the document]
Taluu: I removed the `HttpClient`, even though I'm also not sure due to @dbu's comment... I saw in the PR that introduced http-plug 2.0 (psr-18 compatible) that removing the compatibility for 1.0 was on purpose because it involved some magic.

So I should either target 1.x and thus support also the `HttpClient`, or leave it as is and remove it in here. But, AFAIC, new features should only target 2.0 (master), so....

src/PluginClientBuilder.php
Taluu: I'd rather have to pass the client and the options on the `createClient` for now, especially as we're building a plugin client which has few (one) options.

Adding those setters can be useful indeed, but I don't think it is needed right now, right ?

dbu: i know there is only one option at the moment... but this way is not future proof. if we start adding options, having setOption / removeOption *and* the $options for createClient will be confusing.

src/PluginClientBuilder.php
```diff
@@ -0,0 +1,49 @@
...
+use Psr\Http\Client\ClientInterface;
+
+/**
+ * Build an instance of a PluginClient with a dynamic list of plugins.
+ *
+ * @author Baptiste Clavié <clavie.b@gmail.com>
+ */
+final class PluginClientBuilder
+{
+    /** @var Plugin[][] */
```
dbu: we could say here that this is a map of priority => list of plugins with that priority.

src/PluginClientBuilder.php
```diff
@@ -0,0 +1,49 @@
...
+    private $plugins = [];
+
+    public function addPlugin(Plugin $plugin, int $priority = 0): self
```
dbu: we should add a phpdoc about $priority, to say whether high or low numbers come first in the plugin chain.

src/PluginClientBuilder.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        if (!$client instanceof ClientInterface && !$client instanceof HttpAsyncClient) {
+            throw new \RuntimeException('You must provide a valid http client');
+        }
+
+        $plugins = $this->plugins;
+
+        if (0 === count($plugins)) {
+            $plugins[] = [];
+        }
```
Taluu: note, this is to avoid `array_merge(...[])` which provokes an error. Should I remove the `if` and always add an empty array on the next key when creating the plugin client ? Or should I put it in the properties list `[0 => []]` by default ?)

Taluu: (PHPStan is not happy if I do the inverse condition (`if (count($plugins) > 0) ....`), reporting a possible type violation in the plugin cient's constructor.

dbu: i wonder if we should throw an exception when no plugins are set? if not, i would do `return new PluginClient($client, [], $this->options);` here.

Taluu: If no exceptions are thrown, I don't think it's really worth it to handle this case here if no plugins are given. the `krsort`,`array_merge` and `array_values` on empty array wouldn't really have any impact on perfs, don't you agree ?

dbu: as we need a sepcial case anyways, i think it would be better to be explicit for better readability. 

oh, and thinking about it: throwing an exception would be weird. if we want that, it should be the PluginClient itself, and i guess we do not want it as the builder might sometimes not register any plugins, if we are in some sort of dynamic setup.

GrahamCampbell:
What's the status of this PR?

Taluu:
IIRC, adding tests. I totally forgot about that one. 😂

Taluu:
Test added on the plugins priority check and the fact that the options should be passed onto the plugin client properly

Taluu:
I'm also not in favor of "removing" stuff in a first pass. We could still consider it afterwards though.

dbu:
thanks for the contribution!

johnss:
probably in the future we can add remove plugin,
removing plugin is has many benefit such as clear header that already set
https://github.com/KnpLabs/php-github-api/blob/master/lib/Github/HttpClient/Builder.php#L172
or changing authentication credentials https://github.com/KnpLabs/php-github-api/blob/93c5d19aa1fa214b0cf225acc26f2c8402f0efca/lib/Github/Client.php#L334 if using new instances the request will not be async and lead to lower performance especially for many request (hundred or thousand)

i prefer using unset instead of any other mechanism such as instanceof as it will offer better readability as it does not need "if" statement, you can just use foreach and unset

PR #162, opened 2019-01-20
Minor cleanup

Nyholm:


src/Deferred.php
```diff
@@ -14,15 +14,10 @@
 final class Deferred implements Promise
 {
     private $value;
-
```
dbu: imho it would make sense to have phpdoc for the types of all fields here.

Nyholm: Thank you

src/Deferred.php
```diff
@@ -13,16 +13,34 @@
  */
 final class Deferred implements Promise
 {
+    /**
+     * @var ResponseInterface|null
+     */
     private $value;
 
+    /**
+     * @var Exception|null
```
dbu: should this be Psr\Http\Client\ClientExceptionInterface ? (i see that for now the type declaration in the method is the httplug exception interface, but psr exception would be future proof...

Nyholm: You are correct. Thanks

Nyholm:
Thank you

PR #166, opened 2019-01-21
Update date in changelog

Nyholm:
Minor thing.. 

What do you think about doing an RC-1 after this is merged?


joelwurtz:
Should be merged since it has been released isn't it ?

Nyholm:
I updated the date and sneaked in some doc fixes and cleanup.

Issue #97, opened 2018-03-07
[RFC] Dealing with big/text body


joelwurtz:
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes

#### Actual Behavior

Actually all plugin dealing with body check if the stream is seekable and also made arregments when body is too long via an arbitrary value

#### Expected Behavior

It would be nice to base our default behavior by reading the `Content-Type` header of the request / response and use the https://github.com/php-http/message/blob/master/src/Stream/BufferedStream.php when a body is not seekable

As an example the logger plugin would havec a list of content type authorized for logging (json / text / html / ....) and if stream not seekable, it would be decorated with the `BufferedStream` class

WDYT ? Is something like this considered as a BC Break ?

dbu:
sounds reasonable to me, to increase consistency and reduce the risk of debugging changing how the application works. i guess these will be guidelines and then we apply them to the plugins we manage? or could we code these into a trait that such plugins can use?

i would not consider it a BC break if we do not reduce functionality. if the logger or debug panel sometimes show less information with the changes, i think that would be acceptable.

Issue #45, opened 2016-09-02
Rewrite tests to avoid mocking HTTP messages


sagikazarmark:
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | no
| Version      | <=1.3.x

#### Actual Behavior

Testing our code is pretty hard because of all the mocks necessary to test HTTP Message related code and Plugin Chain.

#### Expected Behavior

It should be pretty easy to understand tests, modify behaviour, provide new test cases, etc.

#### Possible Solutions

- Rewrite specifications to simple unit tests using the MockClient.
- Rewrite specifications to use actual Message objects (so only remove mocking, but preserve specifications)

WDYT?

joelwurtz:
> Rewrite specifications to simple unit tests using the MockClient.

:+1: but would need to do this : https://github.com/php-http/mock-client/issues/3

dbu:
what is the problem, can php-spec not use actual message objects instead of mocks? i'd rather stay consistent with the testing tool if that is not adding another complexity

sagikazarmark:
Of course PHPSpec can use actual objects, but it would be pointless to use the MockClient in that case, since PHPSpec (obviously) does not allow such assertions.

dbu:
nothing is obvious for me with phpspec :-P

should we mix phpunit and phpspec there, or just write some phpunit
tests in addition? we should not overcomplicate things, having one way
to do tests would keep things simple but if phpspec makes it super
complicated to assert things on the MockClient its probably the lesser
evil to have to kind of tests.

joelwurtz:
asserting things with phpspec would make us using phpunit to do assert, so we should go full phpunit IMO

sagikazarmark:
I think using actual message implementations is kind of a midway. But it would not work with mock client.

If we want to use MockClient here, we need PHPUnit, and in that case, I would probable migrate all tests to PHPUnit. Downside is that it's huge amount of work.

But do we really want to use MockClient?

dbu:
if MockClient makes things harder, can we use message implementations
with a phpspec mocked client? i see no reason not to do that if thats
easier.

joelwurtz:
Problem is not the MockClient, if we use actual message implementations we still have to do asserts on the output which will be also message implementations (and not prophecy) so we will have the same problem IMO

fbourigault:
[...]

PR #167, opened 2019-01-21
Add always seekable body plugin

### LINKED ISSUE #98: FilteredStream: unable to read content of stream multiple times

joelwurtz:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #98, replace #99, partially #97  
| Documentation   | https://github.com/php-http/documentation/pull/251
| License         | MIT

#### What's in this PR?

Add a plugin that ensure request body or response body is always seekable

#### Why?

This may be needed when multiple librairies need to acces the content but are not aware of each other, so this allow them to read the body once and then rewind it.


joelwurtz:
Tests are green, don't know why travis ci report this as failed ^^

spec/Plugin/AlwaysSeekableBodyPluginSpec.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        $this->shouldHaveType(AlwaysSeekableBodyPlugin::class);
+    }
+
+    public function it_is_a_plugin()
+    {
+        $this->shouldImplement(Plugin::class);
+    }
+
+    public function it_decorate_response_body_if_not_seekable(RequestInterface $request, ResponseInterface $response, StreamInterface $responseStream, StreamInterface $requestStream)
```
dbu: we have issue #45 saying we should stop mocking request and response objects. i suggest to instantiate actual guzzlehttp/psr7 objects instead.

src/Plugin/AlwaysSeekableBodyPlugin.php
```diff
@@ -0,0 +1,65 @@
...
+
+use Http\Client\Common\Plugin;
+use Http\Message\Stream\BufferedStream;
+use Http\Promise\Promise;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * Decorate the body of the request and the response if it's not seekable by using Http\Message\Stream\BufferedStream.
```
dbu: s/if it's not seekable/if they are not seekable/

dbu:
hm, good question. it could make sense to distinguish that, because i assume adding the buffer is a significant cost that should be avoided when its unnecessary. rather than having flags, i would prefer to split the plugins into SeekableRequestBodyPlugin and SeekableResponseBodyPlugin 

(either way i think we can drop the "Always" from the name, it is more verbose without adding much more clarity imho)

spec/Plugin/AlwaysSeekableBodyPluginSpec.php
joelwurtz: i'm fine with #45 but here i need to test that some method on request / response are being (not) called, so i think it's better to keep the mocking here

dbu: fair enough, then lets keep mocking.

joelwurtz:
As you wish, we can also spli this into 2 plugins : RésponseBodySeekable and RequestBodySeekable, which may be better since you will not put this 2 plugins at the same place : 

 * RequestBodySeekable should be in first so following plugins can seek its body
 * ResponseBodySeekable should be in last so previous plugins will correct have a seekable body on the response

dbu:
didn't even think of that, but yes, lets split the plugin then.

src/Plugin/AlwaysSeekableBodyPlugin.php
joelwurtz: are you sure about that (it = the body) ? However does not matter if splitted :p

dbu: haha, you were right, i got confused. but yes, when we split its going to be singular on both :-)

joelwurtz:
Should be good, will update the documentation accordingly if this is good to merge

PR #169, opened 2019-02-02
Prepare for 1.9.1

Nyholm:
This PR goes to master.


xabbuh:
Shouldn't we merge `1.x` up instead?

Nyholm:
No, they diverged too much. Many CS rules and stuff... Currently I think this is easier.. 

Also, (currently) I dont think there will be many more PRs to 1.x

PR #172, opened 2019-03-15
[Plugin] Add a plugin to record and replay responses

GaryPEGEOT:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | TODO
| License         | MIT

#### What's in this PR?

Add a VCR like plugin to allow storing and replaying response

#### Why?

Allow user to do functional tests why production-like data and provide same functionality as [CsaGuzzleBundle](https://github.com/csarrazi/CsaGuzzleBundle)

#### Example Usage

``` php
<?php
//...
/** @var \Http\Client\Common\Plugin\NamingStrategyInterface $strategy */
$strategy = new MyNamingStrategy()
$plugin = new RecordAndReplayPlugin($strategy, 'some/directory/in/vcs');

// Will perform an HTTP request
$res = $plugin->handleRequest($request, $next, $first)->wait();

// Will be fetch from the FS
$mock = $plugin->handleRequest($request, $next, $first)->wait();

assert($res === $mock);
```

#### Checklist

- [ ] Updated CHANGELOG.md to describe new feature
- [ ] Documentation pull request created

#### To Do

- [ ] provide a default naming strategy
- [ ] make PR to the bundle to provide configuration

**I sadly have no idea how to write PHPSpec tests, so I writed it with PHPUnit, sorry about that...**

### DIFF
--- a/src/Plugin/RecordAndReplayPlugin.php
+++ b/src/Plugin/RecordAndReplayPlugin.php
@@ -1,0 +1,73 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Http\Client\Common\Plugin;
+
+use GuzzleHttp\Psr7;
+use Http\Client\Common\Plugin;
+use Http\Promise\FulfilledPromise;
+use Http\Promise\Promise;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use Symfony\Component\Filesystem\Filesystem;
+
+/**
+ * Record successful responses into the filesystem and replay the response when a similar request is performed (VCR-like).
+ *
+ * @author Gary PEGEOT <garypegeot@gmail.com>
+ */
+final class RecordAndReplayPlugin implements Plugin
+{
+    /**
+     * Return a unique name to identify a given request.
+     *
+     * @var NamingStrategyInterface
+     */
+    private $namingStrategy;
+
+    /**
+     * The directory containing your fixtures (Must be writable).
+     *
+     * @var string
+     */
+    private $directory;
+
+    /**
+     * @var Filesystem
+     */
+    private $fs;
+
+    public function __construct(NamingStrategyInterface $namingStrategy, string $directory, ?Filesystem $fs = null)
+    {
+        $this->namingStrategy = $namingStrategy;
+        $this->directory = $directory;
+        $this->fs = $fs ?? new Filesystem();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first): Promise
+    {
+        if (!$this->fs->exists($this->directory)) {
+            $this->fs->mkdir($this->directory);
+        }
+
+        $directory = realpath($this->directory);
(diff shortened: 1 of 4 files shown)


dbu:
cool, this is interesting. and nice to see how simple it is. when comparing to https://github.com/php-vcr/php-vcr/, i guess "cassettes" would be the realm of the naming strategy (and could be added later, having a simple naming strategy would be enough to get started imho)

the one thing that i would want to have is to have control over whether recording is ok or not. maybe we could split the plugins into RecordPlugin and ReplayPlugin? or rather have a flag to the plugin to tell it how to behave? separate plugins can be used in own code, and in bundle config the DI could decide which plugin to register.

the problem when you don't have that is that you don't notice when you have missing fixtures and the CI will keep doing actual requests.

src/Plugin/RecordAndReplayPlugin.php
```diff
@@ -0,0 +1,73 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Http\Client\Common\Plugin;
+
+use GuzzleHttp\Psr7;
```
dbu: can we read and write requests without a dependency on guzzle? and does that thing handle any response or only guzzle responses?

afaik there is no PSR for (de)serializing responses. if we need to depend on guzzle for it, we should create a separate repository for this plugin that depends on guzzle psr7.

GaryPEGEOT: Sadly it only "deserialize" Guzzle Response, but can serialize any PSR 7 compatible Message. Adding a "mode" (record, replay or both)  is a good idea, I will add it. Adding a new repo would justify the dependency on Guzzzle PSR7 and Symfony's filesystem as well (And PHPUnit in dev), but how can we proceed to create it?

dbu: i can create an empty repository, then you can move the code to a pull request to that repository. like e.g. https://github.com/php-http/logger-plugin

i suggest the name php-http/vcr-plugin. record-and-replay-plugin is too long, and recorder-plugin is incomplete. when we agree on the name, i can create the repo.

GaryPEGEOT: vcr-plugin is fine to me :)

dbu: turns out the https://github.com/php-http/vcr-plugin repository already exists and jerome started something but abandoned it. @sagikazarmark says he talked with jerome and he does not intend to continue.

please do a pull request on it where you completely replace whats in there. jerome started to implement the whole recording and all, but i find your approach much better as there is less code to maintain.

GaryPEGEOT: Perfect! I'll start working on this then

GaryPEGEOT: Done in https://github.com/php-http/vcr-plugin/pull/3

GaryPEGEOT:
Closing as it is duplicated.

PR #183, opened 2020-01-19
adds phpunit as dev dependency

### LINKED ISSUE #182: PHPUnit is missing as dev-dependency and in CI
i was going to analyze a behavior in the library regarding the redirect plugin and was going to utilize phpunit to see whats going on. i was happy to see that there are already some phpunit tests existing that are also relatively new to the project but i had to install phpunit and nyholm/psr7 manually as dev dependency and fix the existing tests a bit to have the chance to get the existing test running. also i found out that the phpunit tests are currently not part of the ci runs.

so my question is: should i try to build a pr that adds the missing dev dependencies and also executes the phpunit tests on ci?

Headd2k:
this pr adds missing phpunit dev dependencies, fixes existing tests and sets base php platform in composer to ensure only compatible dependencies

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fix #182 
| Documentation   | -
| License         | MIT

#### What's in this PR?

- adds required dev dependencies to composer.json to run existing phpunit tests
- fixes existing phpunit tests
- adds phpunit execution to composer test scripts
- adds base platform to composer.json to ensure only php 7.1 compatible dependencies getting added

#### Why?

Existing PHPUnit tests should get executed during CI runs.

#### Checklist

- [x] ~Updated CHANGELOG.md to describe BC breaks / deprecations | new feature | bugfix~ (@dbu thinks: not needed for a testing setup change)

#### To Do

- [x] Need feedback if we could/should replace nyholm/psr7 dependencies/usage in tests with something else
- [x] Need feedback if setting the base platform in composer.json to ensure php 7.1 compatible  dependencies only is okay

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,9 +54,6 @@
         ]
     },
     "config": {
-        "platform": {
-            "php": "7.1.33"
-        },
         "sort-packages": true
     },
     "extra": {
--- a/composer.json
+++ b/composer.json
@@ -20,8 +20,10 @@
     "require-dev": {
         "doctrine/instantiator": "^1.1",
         "guzzlehttp/psr7": "^1.4",
+        "nyholm/psr7": "^1.2",
         "phpspec/phpspec": "^5.1 || ^6.0",
         "phpspec/prophecy": "^1.8",
+        "phpunit/phpunit": "^7.5",
         "sebastian/comparator": "^3.0"
     },
     "suggest": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -42,8 +44,20 @@
...
+            "vendor/bin/phpunit"
+        ],
+        "test-ci": [
+            "vendor/bin/phpspec run -c phpspec.ci.yml",
+            "vendor/bin/phpunit"
+        ]
+    },
+    "config": {
+        "platform": {
+            "php": "7.1.33"
```
dbu: do we need this one? what is it for?

dbu:
sorry, reading the questions now: 

* i would leave the platform thingy out. on travis-ci we do run with all minor php versions we support, that should cover this.
* nyholm/psr7 seems fine to me

Headd2k:
thanks for your quick reply! i updated the pr to remove the base php platform setting. the initial purpose was to have a "safer" process for adding new dependencies. i currently have php 7.2 on my local machine. when i check out the library and run "composer require --dev phpunit/phpunit" i get phpunit 8.x installed because this version is compatible with php 7.2. but the library must support php 7.1. so with the "platform" setting in the composer configuration the require command would have installed phpunit 7.x as this was the last phpunit version branch supporting php 7.1.

dbu:
thanks for the explanation. i don't think we need to worry about adding dependencies: travis-ci will catch it if it goes wrong, and its not often that we add new dependencies. (a library is much more stable than an application usually)

PR #184, opened 2020-02-16
Memory leak hotfix

thomasvargiu:
Hi,
there is a big memory leak when using the `PluginClient` with plugins.

Example code to reproduce the issue:

```php
use Http\Client\Curl\Client;

$samples = 50;
$curlClient = new Client();
$requestFactory = $container->get(\Psr\Http\Message\RequestFactoryInterface::class);
$pluginClient = new \Http\Client\Common\PluginClient(
    $curlClient,
    [
        new \Http\Client\Common\Plugin\RedirectPlugin(),
    ]
);

$request = $requestFactory->createRequest('GET', 'https://google.it');
for ($i = 1; $i <= $samples; $i++) {
    $pluginClient->sendRequest($request);
    echo 'memory: ' . memory_get_usage() . PHP_EOL;
    // gc_collect_cycles(); without this, gc isn't able to free memory
}
```

This PR fixes it removing useless variable reference.

### DIFF
--- a/src/PluginChain.php
+++ b/src/PluginChain.php
@@ -25,11 +25,11 @@ final class PluginChain
     /**
      * @param Plugin[] $plugins
      */
-    public function __construct(array $plugins, callable $clientCallable, int $maxRestarts)
+    public function __construct(array $plugins, callable $clientCallable, array $options = [])
     {
         $this->plugins = $plugins;
         $this->clientCallable = $clientCallable;
-        $this->maxRestarts = $maxRestarts;
+        $this->maxRestarts = (int) ($options['max_restarts'] ?? 0);
     }
 
     private function createChain(): callable
--- a/src/PluginChain.php
+++ b/src/PluginChain.php
@@ -4,10 +4,10 @@ declare(strict_types=1);
 
 namespace Http\Client\Common;
 
+use function array_reverse;
 use Http\Client\Common\Exception\LoopException;
 use Http\Promise\Promise;
 use Psr\Http\Message\RequestInterface;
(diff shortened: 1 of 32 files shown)


GrahamCampbell:
Tests are failing now?

thomasvargiu:
Yes you're right, what I wrote is breaking the chain. I'm going to look at it better

thomasvargiu:
So, the problem is on passing the first callable to plugins, it creates a reference loop.
The only way to resolve it, is to create a new chain of callables for every request.

src/PluginChain.php
```diff
@@ -0,0 +1,58 @@
...
+    /** @var int */
+    private $maxRestarts;
+
+    /** @var int */
+    private $restarts = 0;
+
+    /**
+     * @param Plugin[] $plugins
+     */
+    public function __construct(array $plugins, callable $clientCallable, int $maxRestarts)
```
Nyholm: Lets do `public function __construct(array $plugins, callable $clientCallable, array $options)` instead. That will allow us to add more options without breaking BC

thomasvargiu: Done!

dbu:
> Please revert the CS fixes, then rebase.

@GrahamCampbell is the change with the plugin chain good otherwise?

GrahamCampbell:
Yeh, if you extract just the changes to the files that are needed (I think 3 files?).

GrahamCampbell:
Sorted out a replacement here: https://github.com/php-http/client-common/pull/194.

PR #190, opened 2020-06-30
Support the PSR17 request factory

GrahamCampbell:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

#### What's in this PR?

`Http\Discovery\MessageFactoryDiscovery`  is deprecated, so there is no way to discover a `Http\Message\RequestFactory` instance using non-deprecated APIs. `Http\Client\Common\HttpMethodsClient` should therefore except a PSR17 request factory `Psr\Http\Message\RequestFactoryInterface` as well as a `Http\Message\RequestFactory`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,10 @@
         "php-http/httplug": "^2.0",
         "php-http/message-factory": "^1.0",
         "php-http/message": "^1.6",
-        "symfony/options-resolver": "^2.6 || ^3.4.20 || ~4.0.15 || ~4.1.9 || ^4.2.1 || ^5.0"
+        "psr/http-factory": "^1.0",
+        "psr/http-message": "^1.0",
+        "symfony/options-resolver": "^2.6 || ^3.4.20 || ~4.0.15 || ~4.1.9 || ^4.2.1 || ^5.0",
+        "symfony/polyfill-php80": "^1.17"
     },
     "require-dev": {
         "doctrine/instantiator": "^1.1",
--- a/src/HttpMethodsClient.php
+++ b/src/HttpMethodsClient.php
@@ -6,6 +6,7 @@ namespace Http\Client\Common;
 
 use Http\Message\RequestFactory;
 use Psr\Http\Client\ClientInterface;
+use Psr\Http\Message\RequestFactoryInterface;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
The CI failings are not related to my PR, and can be fixed separately.

composer.json
```diff
@@ -15,6 +15,8 @@
         "php-http/httplug": "^2.0",
         "php-http/message-factory": "^1.0",
         "php-http/message": "^1.6",
+        "psr/http-factory": "^1.0",
+        "psr/http-message": "^1.0",
```
dbu, line 19: why do you add the psr message standard? i don't see any changes in regard to that package?

src/HttpMethodsClient.php
```diff
@@ -17,11 +18,14 @@ final class HttpMethodsClient implements HttpMethodsClientInterface
...
+     * @var RequestFactory|RequestFactoryInterface
      */
     private $requestFactory;
 
-    public function __construct(ClientInterface $httpClient, RequestFactory $requestFactory)
+    /**
+     * @param RequestFactory|RequestFactoryInterface
+     */
+    public function __construct(ClientInterface $httpClient, $requestFactory)
     {
```
dbu, line 29: i would love to have a type check here:

```
if (!$requestFactory instanceof RequestFactory && !$requestFactory instanceof RequestFactoryInterface) {
    throw new \InvalidArgumentException(sprintf('Factory must be of type %s or %s, %s given', RequestFactory::class, RequestFactoryInterface::class, is_object($requestFactory) ? get_class($requestFactory) : gettype($requestFactory)));
}
```

composer.json
```diff
@@ -15,6 +15,8 @@
         "php-http/httplug": "^2.0",
         "php-http/message-factory": "^1.0",
         "php-http/message": "^1.6",
+        "psr/http-factory": "^1.0",
```
dbu, line 18: strictly speaking, a psr factory implementation will require the psr package as well, and when none is available it does not hurt to not have the package installed. but this being a psr with just a few interfaces, i agree to include it for semantic clarity.

composer.json
GrahamCampbell: Strictly speaking, this was not related to this PR, but I included it for the same reason I included the factory: it is directly used by the codebase. I didn't create a separate PR purely because there would be merge conflicts.

src/HttpMethodsClient.php
GrahamCampbell: I've added a check more like what PHP 8 would do if:

image

composer.json
dbu: okay. both are sort of an "optional dependency", but i don't see any harm in pulling in the PSRs. fine for me

src/HttpMethodsClient.php
dbu: awesome, thanks!

GrahamCampbell:
So it turns out this doesn't work quite how we expected! https://github.com/php-http/client-common/pull/202 should fix this.

PR #191, opened 2020-07-01
fix cs config

dbu:
symfony settings added a new default setting to force exceptions on one line. this breaks ci because we do have longer exceptions.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -19,5 +19,6 @@ return PhpCsFixer\Config::create()
         ],
         'no_empty_phpdoc' => true,
         'no_superfluous_phpdoc_tags' => true,
+        'single_line_throw' => false,
     ])
     ->setFinder($finder);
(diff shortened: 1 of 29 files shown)


gmponos:
Things like this will come up again anad again.. unlike PSR2 included in this.. symfony is not a standard hence it's constantly changes.. how about requiring the rules one by one.. 

If you don't have the time right now I can do it later..

dbu:
afaik sometimes changes to cs-fixer itself also lead to suddenly breaking reports, but definitely its more frequent with symfony. and on some things i actually agree, like the ordered imports :-)

i don't have time for it right now, but i would be happy if you want to change it to psr 2 plus as a starting point the rules that symfony currently has. we should do it for all our repos here though, to be consistent... but afaik, it should be possible to really just copy the cs config to all repos. then there might be breaks revealed, so after updating the config it makes sense to run php-cs-fixer on them again.

PR #193, opened 2020-07-01
Consitent implementation of union type checking

GrahamCampbell:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    |yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

This PR consistently implements union type checking. Note that in the cases where there wasn't already a type check in place, the code would have still crashed later, but it is much harder for a programmer to know what went wrong, because it wasn't thrown early enough, where they actually made the made method called, and instead was thrown from deeper in the internals.

### DIFF
--- a/src/PluginClient.php
+++ b/src/PluginClient.php
@@ -50,8 +50,6 @@ final class PluginClient implements HttpClient, HttpAsyncClient
      *
      *     @var int      $max_restarts
      * }
-     *
-     * @throws \RuntimeException if client is not an instance of HttpClient or HttpAsyncClient
      */
     public function __construct($client, array $plugins = [], array $options = [])
     {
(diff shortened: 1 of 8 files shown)


src/PluginClient.php
```diff
@@ -50,8 +50,6 @@ final class PluginClient implements HttpClient, HttpAsyncClient
      *
      *     @var int      $max_restarts
      * }
-     *
-     * @throws \RuntimeException if client is not an instance of HttpClient or HttpAsyncClient
```
GrahamCampbell, line 54: This doc was incorrect anyway, so I think my change is fine, and can be regarded as a bug fix.

NB In general, I don't think we ever need throws docs for "Errors", only "Exceptions", so this is why I've not added any. One can think of Errors in PHP just like how Java's type system treats RuntimeExceptions.

dbu: i agree :+1:

GrahamCampbell:
🚀

PR #194, opened 2020-07-01
Replacement memory leak hotfix

GrahamCampbell:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #184
| License         | MIT

### Original description:

There is a big memory leak when using the PluginClient with plugins.

Example code to reproduce the issue:

```php
use Http\Client\Curl\Client;

$samples = 50;
$curlClient = new Client();
$requestFactory = $container->get(\Psr\Http\Message\RequestFactoryInterface::class);
$pluginClient = new \Http\Client\Common\PluginClient(
    $curlClient,
    [
        new \Http\Client\Common\Plugin\RedirectPlugin(),
    ]
);

$request = $requestFactory->createRequest('GET', 'https://google.it');
for ($i = 1; $i <= $samples; $i++) {
    $pluginClient->sendRequest($request);
    echo 'memory: ' . memory_get_usage() . PHP_EOL;
    // gc_collect_cycles(); without this, gc isn't able to free memory
}
```

This PR fixes it removing useless variable reference.

### DIFF
--- a/src/PluginChain.php
+++ b/src/PluginChain.php
@@ -4,10 +4,10 @@ declare(strict_types=1);
 
 namespace Http\Client\Common;
 
+use function array_reverse;
 use Http\Client\Common\Exception\LoopException;
 use Http\Promise\Promise;
 use Psr\Http\Message\RequestInterface;
-use function array_reverse;
 
 final class PluginChain
 {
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
I'll send a change log PR to add everything at the end if that's ok? This is not the only change worth noting since the last release. :)

dbu:
i usually try to keep the changelog updated constantly, as otherwise these things get forgotten. but with all the parallel things you do that would mean conflicts and rebases after every merge, and i trust you to not forget it at the end :-)

GrahamCampbell:
Actually, this does not need a change log entry, because the fix was made to unrelased code. The plugin chain is not yet tagged. Never the less, we do need a change log entry for the whole feature. I'll sort that out very soon.

PR #202, opened 2020-07-18
Fix broken HttpMethodsClient with PSR RequestFactory

### LINKED ISSUE #201: HttpMethodsClient not working with PSR-17
**Description**
The `HttpMethodsClient` constructor allows to get an `RequestFactory` or `RequestFactoryInterface` as second parameter.

Both declare the `createRequest` but don't have the same argument.

`RequestFactoryInterface::createRequest` only have two parameters `method` and `uri`.

The `HttpMethodsClient::send` method which is used to make the HTTP call have two other parameters. So when using a PSR-17 implementation, the code will not work as the body and/or headers won't be sent.

**How to reproduce**

```php
$http = new HttpMethodsClient(HttpClientDiscovery::find(), Psr17FactoryDiscovery::findRequestFactory());
$response = $http->get($uri, $headers);
```

Headers will not be sent.

GrahamCampbell:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | fixes #201
| License         | MIT

The PSR request factory doesn't know anything about headers or body, and these parameters are silently thrown away!

In order to avoid these, we must set them on the request afterwards. Note that by doing this we have revealed a hidden dependency on a stream factory. I have allowed for this in BC-preserving manner.

### DIFF
--- a/src/HttpMethodsClient.php
+++ b/src/HttpMethodsClient.php
@@ -116,7 +116,7 @@ final class HttpMethodsClient implements HttpMethodsClientInterface
     private function createRequest(string $method, $uri, array $headers = [], $body = null): RequestInterface
     {
         if ($this->requestFactory instanceof RequestFactory) {
-            return = $this->requestFactory->createRequest(
+            return $this->requestFactory->createRequest(
                 $method,
                 $uri,
                 $headers,
--- a/src/HttpMethodsClient.php
+++ b/src/HttpMethodsClient.php
@@ -9,6 +9,9 @@ use Psr\Http\Client\ClientInterface;
 use Psr\Http\Message\RequestFactoryInterface;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\StreamFactoryInterface;
+use Psr\Http\Message\StreamInterface;
+use Psr\Http\Message\UriInterface;
 
 final class HttpMethodsClient implements HttpMethodsClientInterface
 {
@@ -23,9 +26,14 @@ final class HttpMethodsClient implements HttpMethodsClientInterface
     private $requestFactory;
 
     /**
+     * @var StreamFactoryInterface|null
+     */
+    private $streamFactory;
+
+    /**
      * @param RequestFactory|RequestFactoryInterface
      */
-    public function __construct(ClientInterface $httpClient, $requestFactory)
+    public function __construct(ClientInterface $httpClient, $requestFactory, StreamFactoryInterface $streamFactory = null)
     {
         if (!$requestFactory instanceof RequestFactory && !$requestFactory instanceof RequestFactoryInterface) {
             throw new \TypeError(
@@ -33,8 +41,13 @@ final class HttpMethodsClient implements HttpMethodsClientInterface
             );
         }
 
+        if (!$requestFactory instanceof RequestFactory && null === $streamFactory) {
+            @trigger_error(sprintf('Passing a %s without a %s to %s::__construct() is deprecated as of version 2.3 and will be disallowed in version 3.0. A stream factory is required to create a request with a string body.', RequestFactoryInterface::class, StreamFactoryInterface::class, self::class));
+        }
+
         $this->httpClient = $httpClient;
         $this->requestFactory = $requestFactory;
+        $this->streamFactory = $streamFactory;
     }
 
     public function get($uri, array $headers = []): ResponseInterface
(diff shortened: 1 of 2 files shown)


src/HttpMethodsClient.php
```diff
@@ -79,12 +92,55 @@ public function options($uri, array $headers = [], $body = null): ResponseInterf
...
+            return $this->requestFactory->createRequest(
+                $method,
+                $uri,
+                $headers,
+                $body
+            );
+        }
+
+        if (is_string($body) && null === $this->streamFactory) {
+            throw new \RuntimeException('Cannot create request: A stream factory is required to create a request with a string body.');
```
dbu, line 128: the strict BC way would be to ignore the body, right? though that very much sounds like a bug, so the "BC break" is that you now realize that your code is not working as expected.

but we should also ignore `''` then, to avoid unnecessarily breaking systems that happened to send an empty string.

src/HttpMethodsClient.php
```diff
@@ -22,19 +25,29 @@ final class HttpMethodsClient implements HttpMethodsClientInterface
...
+    public function __construct(ClientInterface $httpClient, $requestFactory, StreamFactoryInterface $streamFactory = null)
     {
         if (!$requestFactory instanceof RequestFactory && !$requestFactory instanceof RequestFactoryInterface) {
             throw new \TypeError(
                 sprintf('%s::__construct(): Argument #2 ($requestFactory) must be of type %s|%s, %s given', self::class, RequestFactory::class, RequestFactoryInterface::class, get_debug_type($requestFactory))
             );
         }
 
+        if (!$requestFactory instanceof RequestFactory && null === $streamFactory) {
+            @trigger_error(sprintf('Passing a %s without a %s to %s::__construct() is deprecated as of version 2.3 and will be disallowed in version 3.0. A stream factory is required to create a request with a string body.', RequestFactoryInterface::class, StreamFactoryInterface::class, self::class));
```
dbu, line 45: i assume that in next major, we might want to only support the psr factory, to move forward? then we could simply make the stream factory a required argument... (no need to change anything here, just a thought ;-) )

src/HttpMethodsClient.php
GrahamCampbell: Yeh, we could ignore the empty string too.

src/HttpMethodsClient.php
GrahamCampbell: I think so too. I am thinking once the old interfaces are all marked as deprecated in the future (https://github.com/php-http/message-factory/blob/master/src/ResponseFactory.php doesn't have a deprecation annotation yet), we can come back here to this constructor and also add a futher deprecation warning for if a psr request factory is not passed. Then yeh, in the next major version make all prams typed with psr interfaces, and all required.

src/HttpMethodsClient.php
```diff
@@ -79,12 +92,55 @@ public function options($uri, array $headers = [], $body = null): ResponseInterf
 
     public function send(string $method, $uri, array $headers = [], $body = null): ResponseInterface
     {
-        return $this->sendRequest($this->requestFactory->createRequest(
-            $method,
-            $uri,
-            $headers,
-            $body
-        ));
+        if (!is_string($uri) && !$uri instanceof UriInterface) {
```
GrahamCampbell, line 95: Note that this is not a BC break, even if the caller didn't have strict typing and expected stuff to be cast to a string at runtime, because this file itself has strict typing and was previously forwarding the parameter as is, so the result would still be a type error later, but just a more obscure error message. Including an early guard here will help callers correct their error, and also checking the uri and body are the correct types here mean our logic later on within the private createRequest function is a little simpler than it would have otherwise been.

GrahamCampbell:
Good call about the empty string. I've allowed that case through without setting the body. :)

src/HttpMethodsClient.php
GrahamCampbell: Done!

GrahamCampbell:
After this is merged, don't tag a release yet. I am going to add phpstan to the CI in a separate PR, and see if I can detect any further bugs.

dbu:
thanks!

dbu:
phpstan is a good idea for sure. i don't know if it would have spotted this particular issue, because the factory can have one or the other type, and the call was correct for one type.

GrahamCampbell:
It would have warned us that we could possibly be passing too many parameters, if we had the analysis level turned up high enough (actually, level 2 or higher). It'd have insisted we did an instanceof check:

image

image

PR #203, opened 2020-07-19
Static analysis fixes

GrahamCampbell:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

PHPStan has detected 3 more bugs for us, ~and some unreachable code in the plugin client~.

1. Fixed bug in the cookie plugin
2. Fixed null bug in when parsing date header
3. Fixed typo in methods client phpdoc
4. ~Removed unreachable plugin client code~
5. Don't temprarily violate property types in the flexible http client
6. Added phpstan and some extra type information to help it, in some places


src/PluginClientFactory.php
```diff
@@ -28,8 +28,10 @@ final class PluginClientFactory
      * application execution.
      *
      * @internal
+     *
+     * @param callable(ClientInterface|HttpAsyncClient, Plugin[], array): PluginClient $factory
      */
-    public static function setFactory(callable $factory)
+    public static function setFactory(callable $factory): void
```
GrahamCampbell, line 34: Not a BC break: the class is final.

src/Plugin/CookiePlugin.php
```diff
@@ -167,13 +167,13 @@ private function createCookie(RequestInterface $request, string $setCookieHeader
...
      *
-     * @return string[]
+     * @return array{0:string, 1:?string}
      */
     private function createValueKey(string $part): array
     {
         $parts = explode('=', $part, 2);
         $key = trim($parts[0]);
-        $value = isset($parts[1]) ? trim($parts[1]) : true;
+        $value = isset($parts[1]) ? trim($parts[1]) : null;
```
GrahamCampbell, line 176: Not a BC break: the Cookie class only accepts string or null as a value. This is a bug fix.

src/HttpClientPool/HttpClientPool.php
```diff
@@ -25,13 +25,14 @@ abstract class HttpClientPool implements HttpClientPoolInterface
     /**
      * Add a client to the pool.
      *
-     * @param ClientInterface|HttpAsyncClient|HttpClientPoolItem $client
```
GrahamCampbell, line 28: Not a BC break. A `HttpClientPoolItem` is a `ClientInterface`. It's incorrect to specify it again.

GrahamCampbell:
It looks like that dead was code intentionally in there, even though it was not needed...

phpstan.neon.dist
```diff
@@ -0,0 +1,31 @@
+parameters:
+    level: max
+    checkMissingIterableValueType: false
+    treatPhpDocTypesAsCertain: false
+    paths:
+        - src
+    ignoreErrors:
+        -
+            message: "#^Strict comparison using \\!\\=\\= between null and null will always evaluate to false\\.$#"
```
GrahamCampbell, line 9: This is a false positive, and should be reported to phpstan as a bug.

phpstan.neon.dist
```diff
@@ -0,0 +1,31 @@
...
+            count: 1
+            path: src/HttpMethodsClient.php
+
+        -
+            message: "#^Cannot call method createStream\\(\\) on Psr\\\\Http\\\\Message\\\\StreamFactoryInterface\\|null\\.$#"
```
GrahamCampbell, line 14: This is a false positive, and should be reported to phpstan as a bug.

phpstan.neon.dist
```diff
@@ -0,0 +1,31 @@
...
+            count: 1
+            path: src/HttpMethodsClient.php
+
+        -
+            message: "#^Method Http\\\\Client\\\\Common\\\\Plugin\\\\Journal\\:\\:addSuccess\\(\\) has no return typehint specified\\.$#"
```
GrahamCampbell, line 19: This is a no-fix. Changing it would break BC. Maybe we should consider adding void return to the interface in the next major version.

phpstan.neon.dist
```diff
@@ -0,0 +1,31 @@
...
+            count: 1
+            path: src/Plugin/Journal.php
+
+        -
+            message: "#^Method Http\\\\Client\\\\Common\\\\Plugin\\\\Journal\\:\\:addFailure\\(\\) has no return typehint specified\\.$#"
```
GrahamCampbell, line 24: This is a no-fix. Changing it would break BC. Maybe we should consider adding void return to the interface in the next major version.

phpstan.neon.dist
```diff
@@ -0,0 +1,31 @@
...
+            count: 1
+            path: src/Plugin/Journal.php
+
+        -
+            message: "#^Call to an undefined method Http\\\\Client\\\\HttpAsyncClient\\:\\:sendRequest\\(\\)\\.$#"
```
GrahamCampbell, line 29: This is a false positive, and should be reported to phpstan as a bug.

dbu:
> Removed unreachable plugin client code

are we missing some tests with that code?

GrahamCampbell:
> are we missing some tests with that code?

No, there are tests, and that's what made me put the code back.

GrahamCampbell:
Maybe unreachable was the wrong word. It's more that the code was duplicated, however that seems to have been intentional.

Issue #205, opened 2020-09-23
RedirectPlugin sends body in GET requests


josiasmontag:
**Description**

If the initial request method is `POST` and the server responds with a `Location` header, the `RedirectPlugin` is configured to follow the redirect with a `GET` request. However, the `RedirectPlugin` will resend the body of the initial request. This results in a `GET` request with a request body.

Most servers will just ignore the request body of `GET` requests. Nevertheless, this causes trouble:
* AWS CloudFront rejects GET requests with request body
* Unnecessary traffic / performance issues

**Possible Solution**

The RedirectPlugin clones the original request and changes the method to `GET` at this point: 
https://github.com/php-http/client-common/blob/e37e46c610c87519753135fb893111798c69076a/src/Plugin/RedirectPlugin.php#L187

When switching to `GET` it should also remove the body. 

However, I do not see any way to remove the body easily. One quick & dirty solution that worked for me is to use an empty stream: 
```php
 $originalRequest = $originalRequest->withBody(GuzzleHttp\Psr7\stream_for(''));
```

dbu:
hm, i imagine this could also have security implications, when sending a body to an unexpected target.

on the other hand, it really depends. GET request with body are not prohibited by HTTP. elasticsearch for example uses them to query the index (though they also allow POST, to get around limitations in the network like the one you describe).

i suggest that we introduce a configuration option in the redirect plugin `drop_body_when_change_to_get` or something like that. if we only drop to an empty body when the redirect changes the method (we enter https://github.com/php-http/client-common/blob/e37e46c610c87519753135fb893111798c69076a/src/Plugin/RedirectPlugin.php#L187 and the method actually is different).

which leaves the question of *how* to empty the body. short of providing the stream factory as the value of the option, i don't see how it can be done, but maybe i am missing something.

dbu:
@Nyholm do you have an idea how we can empty the body of the request?

GrahamCampbell:
>  $originalRequest = $originalRequest->withBody(GuzzleHttp\Psr7\stream_for(''));

The problem with doing that is other headers may now be wrong, such as the content type header.

dbu:
hm, that as well. guess we should also remove the content-type request header when changing to a GET request. content-length should also go. can we mimik the behaviour of guzzle, or is that also not completely correct?

a dependency on guzzle would be a problem - the client-common package afaik does not depend on a concrete implementation. injecting the stream factory sounds like a lot of complication, but would probably be the correct thing.

[...]

PR #208, opened 2021-07-02
Added `strict` option for RedirectPlugin not to modify request method…

### LINKED ISSUE #207: Add an option for RedirectPlugin not to modify request on statuses 300, 301, 302
**Description**
I doubted if I need to create a bug ticket, but then as I also will follow up with a solution that in backwards compatible, I would like to post the issue as a feature request

Currently statuses 300, 301, 302 are switching from methods that are not GET and HEAD to GET, even though specification states that it's not allowed, but at the same time, it is a common implementation by http clients and browsers

https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3

> If the 302 status code is received in response to a request other than GET or HEAD, the user agent MUST NOT automatically redirect the request unless it can be confirmed by the user, since this might change the conditions under which the request was issued.
> 
> Note: RFC 1945 and RFC 2068 specify that the client is not allowed
> to change the method on the redirected request. However, most
> existing user agent implementations treat 302 as if it were a 303
> response, performing a GET on the Location field-value regardless
> of the original request method. The status codes 303 and 307 have
> been added for servers that wish to make unambiguously clear which
> kind of reaction is expected of the client.

So, according to specs that should happen

Request 1. POST http://example.com/api
Response 1. 302 Location https://example.com/api
Request 2. POST https://example.com/api

Currently that happens with client-common's RedirectPlugin

Request 1. POST http://example.com/api
Response 1. 302 Location https://example.com/api
Request 2. GET https://example.com/api

My suggestion would be to add `strict` option with default value `false` to RedirectPlugin a very similar way as guzzle's RedirectMiddleware does that:

https://github.com/guzzle/guzzle/blob/de6f1e58e735754b888649495ed4cb9ae3b19589/src/RedirectMiddleware.php#L157-L169
```
        // Use a GET request if this is an entity enclosing request and we are
        // not forcing RFC compliance, but rather emulating what all browsers
        // would do.
        $statusCode = $response->getStatusCode();
        if ($statusCode == 303 ||
            ($statusCode <= 302 && !$options['allow_redirects']['strict'])
        ) {
            $safeMethods = ['GET', 'HEAD', 'OPTIONS'];
            $requestMethod = $request->getMethod();


            $modify['method'] = in_array($requestMethod, $safeMethods) ? $requestMethod : 'GET';
            $modify['body'] = '';
        }
```

**Additional context**
* Related issue, but it's about sending body after the method was updated https://github.com/php-http/client-common/issues/205
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301

> Even if the specification requires the method (and the body) not to be altered when the redirection is performed, not all user-agents align with it - you can still find this type of bugged software out there. It is therefore recommended to use the 301 code only as a response for GET or HEAD methods and to use the 308 Permanent Redirect for POST methods instead, as the method change is explicitly prohibited with this status.
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302

> Even if the specification requires the method (and the body) not to be altered when the redirection is performed, not all user-agents conform here - you can still find this type of bugged software out there. It is therefore recommended to set the 302 code only as a response for GET or HEAD methods and to use 307 Temporary Redirect instead, as the method change is explicitly prohibited in that case.
> 
> In the cases where you want the method used to be changed to GET, use 303 See Other instead. This is useful when you want to give a response to a PUT method that is not the uploaded resource but a confirmation message such as: 'you successfully uploaded XYZ'.

* I could not find if method and body are not meant to be unmodified when 300, I propose implementing it the same way as guzzle has it - also strict

olexiyk:
… on statuses 300, 301, 302

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #207, partially #205
| License         | MIT

#### What's in this PR?

Resolves #207 by allowing to follow redirect specs for 301 and 302 more strictly - not modifying request method 

#### Example Usage

``` php
        $plugins = $plugins ?? [
            new RedirectPlugin(['strict' => true]),
        ];

        $httpClient = (new PluginClientFactory())->createClient(
            Psr18ClientDiscovery::find(),
            $plugins
        );
```

#### Checklist

#### To Do

- [ ] If the PR is not complete but you want to discuss the approach, list what remains to be done here

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 
-- `strict` option to `RedirectPlugin`
+- `strict` option to `RedirectPlugin` to allow preserving the request method on redirections with status 300, 301 and 302.
 
 ## 2.3.0 - 2020-07-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## 2.4.0 (unreleased)
+
+### Added
+
+- `strict` option to `RedirectPlugin`
+
 ## 2.3.0 - 2020-07-21
 
 ### Fixed
@@ -112,7 +118,7 @@
 
 ## 1.7.0 - 2017-11-30
 
-### Added 
+### Added
 
 - Symfony 4 support
 
@@ -132,12 +138,12 @@
 
 ### Changed
 
-- The `RetryPlugin` does now wait between retries. To disable/change this feature you must write something like: 
- 
+- The `RetryPlugin` does now wait between retries. To disable/change this feature you must write something like:
+
 ```php
-$plugin = new RetryPlugin(['delay' => function(RequestInterface $request, Exception $e, $retries) { 
-  return 0; 
-}); 
+$plugin = new RetryPlugin(['delay' => function(RequestInterface $request, Exception $e, $retries) {
+  return 0;
+});
 ```
 
 ### Deprecated
--- a/src/Plugin/RedirectPlugin.php
+++ b/src/Plugin/RedirectPlugin.php
@@ -105,7 +105,7 @@ final class RedirectPlugin implements Plugin
      * @param array $config {
      *
      *     @var bool|string[] $preserve_header True keeps all headers, false remove all of them, an array is interpreted as a list of header names to keep
-     *     @var bool $use_default_for_multiple Whether the location header must be directly used for a multiple redirection status code (300).
+     *     @var bool $use_default_for_multiple Whether the location header must be directly used for a multiple redirection status code (300)
      *     @var bool $strict When true, redirect codes 300, 301, 302 will not modify request method and body.
      * }
      */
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+## 2.4.0 (unreleased)
+
+### Added
+
+- `strict` option to `RedirectPlugin`
```
dbu, line 7:
```suggestion
- `strict` option to `RedirectPlugin` to allow preserving the request method on redirections with status 300, 301 and 302.
```

dbu (approved):
thanks for the pull request and the detailed explanation. this makes sense to me :+1: 

can you update https://github.com/php-http/documentation/blob/master/plugins/redirect.rst for the documentation please?

src/Plugin/RedirectPlugin.php
```diff
@@ -106,6 +106,7 @@ final class RedirectPlugin implements Plugin
      *
      *     @var bool|string[] $preserve_header True keeps all headers, false remove all of them, an array is interpreted as a list of header names to keep
      *     @var bool $use_default_for_multiple Whether the location header must be directly used for a multiple redirection status code (300).
```
dbu, line 108:
```suggestion
     *     @var bool $use_default_for_multiple Whether the location header must be directly used for a multiple redirection status code (300)
```

pretty-ci has weird rules on this construct, but lets keep it green

dbu (approved):
thanks again!

dbu:
i tagged 2.4.0 with this feature. thanks for the contribution!

olexiyk:
Thank you for quick action!

PR #215, opened 2021-11-23
Dependency: Symfony/options-resolver allow ^6.0

Jeroeny:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | None
| License         | MIT

#### What's in this PR?

Update `composer.json` to allow the upcoming `^6.0` for `symfony/options-resolver`.

#### Why?

To support Symfony 6.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -66,7 +66,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        symfony: ['2', '3', '4']
+        symfony: ['2', '3', '4', '5', '6']
 
     steps:
       - name: Checkout code
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -66,7 +66,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        symfony: ['2', '3', '4', '5', '6']
+        symfony: ['4', '5', '6']
 
     steps:
       - name: Checkout code
@@ -85,8 +85,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          composer require dunglas/symfony-lock:v${{ matrix.symfony }} --no-interaction --no-update
-          composer update --prefer-dist --no-interaction --prefer-stable --prefer-lowest --no-progress
+          SYMFONY_REQUIRE=${{ matrix.symfony }} composer update --prefer-dist --no-interaction --prefer-stable --prefer-lowest --no-progress
 
       - name: Execute tests
         run: composer test
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -85,7 +85,9 @@ jobs:
 
       - name: Install dependencies
         run: |
-          SYMFONY_REQUIRE=${{ matrix.symfony }} composer update --prefer-dist --no-interaction --prefer-stable --prefer-lowest --no-progress
+          composer require --no-update --no-interaction --no-progress symfony/flex
+          composer config extra.symfony.require ${{ matrix.symfony}}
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
Do the tests pass against symfony 6?

Jeroeny:
> Do the tests pass against symfony 6?

The result is: `Tests: 14, Assertions: 24, Warnings: 6.`, if that was the baseline, then yes.

Output: https://gist.github.com/Jeroeny/acc1ac1fe61927d3cf8951d6310d12a2

GrahamCampbell (approved):

dbu:
thanks!

could you add symfony 6 (and 5 :scream: ) to the build matrix at .github/workflows/tests.yml please?

Jeroeny:
@dbu 👍 added

.github/workflows/tests.yml
```diff
@@ -66,7 +66,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        symfony: ['2', '3', '4']
+        symfony: ['2', '3', '4', '5', '6']
```
GrahamCampbell, line 69: This alone won't work. You'll need to exclude symfony 5 on php 7.1, and symfony 6 on all php 7.x versions

GrahamCampbell (changes requested):

.github/workflows/tests.yml
GrahamCampbell: It also won't work because no such version of `dunglas/symfony-lock` exists.

GrahamCampbell (changes requested):

.github/workflows/tests.yml
dbu: oops, sorry jeroeny.

@GrahamCampbell my thought would be to drop support for symfony 2 and 3 (those are super old, and people stuck on those versions can still use the existing releases of symfony) and change the build to use flex to lock the symfony version instead of dunglas/symfony-lock. do you agree?

GrahamCampbell: sounds good.

Jeroeny: is the environment variable `SYMFONY_REQUIRE` used to lock the Symfony version in Flex? See new commit

.github/workflows/tests.yml
```diff
@@ -85,8 +85,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          composer require dunglas/symfony-lock:v${{ matrix.symfony }} --no-interaction --no-update
-          composer update --prefer-dist --no-interaction --prefer-stable --prefer-lowest --no-progress
+          SYMFONY_REQUIRE=${{ matrix.symfony }} composer update --prefer-dist --no-interaction --prefer-stable --prefer-lowest --no-progress
```
dbu, line 88: afaik we need a line before this one like

            composer require --no-update symfony/flex && composer config extra.symfony.require ${{ matrix.symfony}}

.github/workflows/tests.yml
dbu: thanks a lot for also handling the ci setup, jeroeny

.github/workflows/tests.yml
Jeroeny: Seems to work at least locally. I've committed the changes.

dbu:
thanks!

there seems to have been a change with phpstan but that is definitely not related to your changes. i will merge and look at the phpstan issues separately.

dbu:
thanks again @Jeroeny , i tagged https://github.com/php-http/client-common/releases/tag/2.5.0 with symfony 6 support.

PR #216, opened 2021-11-25
cleanups found with phpstan

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | mentioned in #215
| Documentation   | -
| License         | MIT

#### What's in this PR?

Adjust to new phpstan version, cleanup some phpdoc

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,8 +13,8 @@
     "require": {
         "php": "^7.1 || ^8.0",
         "php-http/httplug": "^2.0",
-        "php-http/message-factory": "^1.0",
         "php-http/message": "^1.6",
+        "php-http/message-factory": "^1.0",
         "psr/http-client": "^1.0",
         "psr/http-factory": "^1.0",
         "psr/http-message": "^1.0",
--- a/src/Plugin/RetryPlugin.php
+++ b/src/Plugin/RetryPlugin.php
@@ -56,14 +56,14 @@ final class RetryPlugin implements Plugin
     private $retryStorage = [];
 
     /**
-     * @param array $config {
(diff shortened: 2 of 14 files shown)


composer.json
```diff
@@ -13,8 +13,9 @@
     "require": {
         "php": "^7.1 || ^8.0",
         "php-http/httplug": "^2.0",
-        "php-http/message-factory": "^1.0",
         "php-http/message": "^1.6",
+        "php-http/message-factory": "^1.0",
+        "phpstan/phpstan": "^1.2",
```
Nyholm, line 18: Oops :)

src/Plugin/RetryPlugin.php
```diff
@@ -140,6 +141,7 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
             }
 
             $time = call_user_func($this->exceptionDelay, $request, $exception, $this->retryStorage[$chainIdentifier]);
+            \assert(is_int($time));
```
Nyholm, line 144: Will this benefit the user?
If it is just to convince phpstan we can add comments.

Nyholm (changes requested):
Thank you. 

It looks great. Just remove phpstan from composer.json and I’ll be happy.

src/Plugin/RetryPlugin.php
GrahamCampbell: Better to add a `/**@var int */` or whatever. Agreed.

composer.json
dbu: oh, oops indeed. will remove it again.

src/Plugin/RetryPlugin.php
dbu: this way the user gets a warning if the user func returns the wrong thing.

but i can also put the annotation.

PR #218, opened 2022-09-10
RedirectPlugin: Default to empty path when Location doesn't specify any

### LINKED ISSUE #217: RedirectPlugin is wrongly overwriting path and falsely detects circular redirect
**PHP version**: 8.1.7

**Description**
Given HTTPlug requests http://foo/entry?deviceoutput=app
And URL above responds with redirect to http://foo?deviceoutput=app
Then RedirectPlugin falsely detects circular redirect

**Possible Solution**
I believe reason for this is because original request is used as a base object for target location here https://github.com/php-http/client-common/blob/master/src/Plugin/RedirectPlugin.php#L233

And path from it is not overridden if it's not specified in location header.

ostrolucky:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #217
| Documentation   |
| License         | MIT

### DIFF
--- a/src/Plugin/RedirectPlugin.php
+++ b/src/Plugin/RedirectPlugin.php
@@ -231,6 +231,11 @@ final class RedirectPlugin implements Plugin
         }
 
         $uri = $originalRequest->getUri();
+        $uri = $uri
+            ->withPath(array_key_exists('path', $parsedLocation) ? $parsedLocation['path'] : '')
+            ->withQuery(array_key_exists('query', $parsedLocation) ? $parsedLocation['query'] : '')
+            ->withFragment(array_key_exists('fragment', $parsedLocation) ? $parsedLocation['fragment'] : '')
+        ;
 
         if (array_key_exists('scheme', $parsedLocation)) {
             $uri = $uri->withScheme($parsedLocation['scheme']);
@@ -242,22 +247,6 @@ final class RedirectPlugin implements Plugin
 
         if (array_key_exists('port', $parsedLocation)) {
             $uri = $uri->withPort($parsedLocation['port']);
-        }
-
-        if (array_key_exists('path', $parsedLocation)) {
-            $uri = $uri->withPath($parsedLocation['path']);
-        }
-
-        if (array_key_exists('query', $parsedLocation)) {
-            $uri = $uri->withQuery($parsedLocation['query']);
-        } else {
-            $uri = $uri->withQuery('');
-        }
-
-        if (array_key_exists('fragment', $parsedLocation)) {
-            $uri = $uri->withFragment($parsedLocation['fragment']);
-        } else {
-            $uri = $uri->withFragment('');
         }
 
         return $uri;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## 2.5.1 - 2022-09-??
+
+### Fixed
+
+- Fixes false positive circular detection in RedirectPlugin in cases when target location does not contain path
+
 ## 2.5.0 - 2021-11-26
 
 ### Added
--- a/spec/Plugin/RedirectPluginSpec.php
+++ b/spec/Plugin/RedirectPluginSpec.php
@@ -162,6 +162,9 @@ class RedirectPluginSpec extends ObjectBehavior
 
         $request->getUri()->willReturn($uri);
         $uri->withScheme('https')->willReturn($uriRedirect);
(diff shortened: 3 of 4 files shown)


src/Plugin/RedirectPlugin.php
```diff
@@ -244,22 +249,6 @@ private function createUri(ResponseInterface $redirectResponse, RequestInterface
             $uri = $uri->withPort($parsedLocation['port']);
```
dbu, line 249: i am actually not sure about this one either. if i was on http://localhost:8080/foo and i get redirect a header https://php-http.org/bar i would not expect to stay on port 8080... the only case where the port should be kept is when there is also no host in the redirection... wdyt?

dbu:
thanks for looking into this! i think you are right that we need to also reset the path if none is provided, and i like how you compact the code by chaining the calls.

can you please check the test failure? is this some mocking going wrong? could we use an actual URI implementation instead of mocking, if thats the cause for the issue?

src/Plugin/RedirectPlugin.php
ostrolucky: I agree. Will change

ostrolucky:
Wow those phpspec tests are insane. Can I just delete those phpspec tests I have trouble with, since I replaced it with integration-like new phpunit test? Otherwise it would take me entire day to fix it, as phpspcec is makint it very difficult to find what needs to be changed.

dbu:
i guess the problem is that the Uri object is mocked instead of using an actual implementation, and it is missing mocking those calls. but i agree to remove the failing cases in favor of the phpunit tests, i also find them a lot more readable. please add a phpdoc to the top of the RedirectPluginSpec that mentions the phpunit test and vice versa.

dbu:
@ostrolucky do you remove the failing phpspec tests and also fix the thing about the port?

ostrolucky:
Yes I just didn't have time to continue since my last struggle, but I have unstaged changes

ostrolucky:
Should be addressed now.

src/Plugin/RedirectPlugin.php
```diff
@@ -230,36 +230,18 @@ private function createUri(ResponseInterface $redirectResponse, RequestInterface
...
-
-        if (array_key_exists('scheme', $parsedLocation)) {
-            $uri = $uri->withScheme($parsedLocation['scheme']);
-        }
+        $uri = $originalRequest->getUri()
+            ->withPath(array_key_exists('path', $parsedLocation) ? $parsedLocation['path'] : '')
+            ->withQuery(array_key_exists('query', $parsedLocation) ? $parsedLocation['query'] : '')
+            ->withFragment(array_key_exists('fragment', $parsedLocation) ? $parsedLocation['fragment'] : '')
+            ->withScheme(array_key_exists('scheme', $parsedLocation) ? $parsedLocation['scheme'] : 'http')
+            ->withPort(array_key_exists('port', $parsedLocation) ? $parsedLocation['port'] : null)
```
dbu, line 238: i think scheme and port can not be handled like this. if there was no domain in the redirect at all, we should keep scheme and port from the original domain.

ostrolucky: I thought you agreed on this when you commented this
> i am actually not sure about this one either. if i was on http://localhost:8080/foo and i get redirect a header [php-http.org/bar](https://php-http.org/bar) i would not expect to stay on port 8080... the only case where the port should be kept is when there is also no host in the redirection... wdyt?

ostrolucky: I think taking over scheme/port will cause same class of bug for somebody else as I had with path. If hosts change, HTTP client should not use same mechanism as browser does when hosts don't change.

dbu: yes, we agreed on that. but there are more cases:
* `Location http://otherhost/foobar` => drop the port.
* `Location ://otherhost/foobar` => drop port, keep scheme
* `Location `/foobar` => do not touch scheme, host, port

dbu: fair enough. sorry for the feature creep. i created https://github.com/php-http/client-common/issues/219

dbu:
i fixed the build issue in #220, can you please rebase this branch? i don't expect problems with only one symfony version, but it just feels too yolo to merge a PR with failing builds :-)

ostrolucky:
rebased

dbu:
yay, thanks!

PR #221, opened 2022-09-27
redirection to different domain must not keep previous port

### LINKED ISSUE #219: RedirectPlugin port handling
**Description**

As discussed in #218, there are some cases of redirections where our plugin will mess up. Currently we only overwrite the parts that are redefined. These cases will not work correctly when starting with e.g. `https://localhost:8000`:

* `Location: http://otherhost/foobar` => need to drop the port (and we already replace schema and domain)
* `Location ://otherhost/foobar` => need to drop port (and we already keep scheme which happens to be correct, and replace domain)
* `Location: /foobar` => already correct, we don't touch any of the domain, scheme or port

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #219
| Documentation   | -
| License         | MIT

#### What's in this PR?

Correctly handle non-absolute URLs in the redirect

#### Why?

Conform with [RFC 7231 Location header specification](https://www.rfc-editor.org/rfc/rfc7231#section-7.1.2) 

#### Checklist


tests/Plugin/RedirectPluginTest.php
```diff
@@ -26,10 +26,26 @@ function () {}
...
+            'relative-path with ./' => ["https://example.com:8000/path/", "./other?query=value", "https://example.com:8000/path/other?query=value"],
+            'relative-path with //' => ["https://example.com:8000/path/", "other//sub?query=value", "https://example.com:8000/path/other//sub?query=value"],
+            'relative-path redirect with only query' => ["https://example.com:8000/path", "?query=value", "https://example.com:8000/path?query=value"],
+       ];
+    }
+
     /**
-     * @testWith ["https://example.com/path?query=value", "https://example.com?query=value", "https://example.com?query=value"]
-     *           ["https://example.com/path?query=value", "https://example.com/?query=value", "https://example.com/?query=value"]
-     *           ["https://example.com", "https://example.com?query=value", "https://example.com?query=value"]
```
dbu, line 32: moved to provider method to have descriptions of the cases

dbu:
@ostrolucky can you review this one please? trying to get all of the redirection right. i discovered that besides the domain/scheme/port topic, RFC 7231 also allows relative paths with all of the path semantics...

src/Plugin/RedirectPlugin.php
```diff
@@ -231,8 +231,34 @@ private function createUri(ResponseInterface $redirectResponse, RequestInterface
...
+            if ('' === $path) {
+                $path = $originalPath;
+            } elseif (($pos = strrpos($originalPath, '/')) !== false) {
+                $path = substr($originalPath, 0, $pos+1).$path;
+            } else {
+                $path = '/'.$path;
+            }
+            /* replace '/./' or '/foo/../' with '/' */
+            $re = ['#(/\./)#', '#/(?!\.\.)[^/]+/\.\./#'];
+            for( $n=1; $n>0; $path=preg_replace( $re, '/', $path, -1, $n ) ) {
```
ostrolucky, line 251: you have som weird code style here. Need to be a bit more careful because this project doesn't have cs check

src/Plugin/RedirectPlugin.php
```diff
@@ -231,8 +231,34 @@ private function createUri(ResponseInterface $redirectResponse, RequestInterface
         }
 
         $uri = $originalRequest->getUri();
+
+        // Redirections can either use an absolute uri or a relative reference https://www.rfc-editor.org/rfc/rfc3986#section-4.2
+        // If relative, we need to check if we have an absolute path or not
+
+        $path = array_key_exists('path', $parsedLocation) ? $parsedLocation['path'] : '';
+        if (!array_key_exists('host', $parsedLocation) && '/' !== $location[0]) {
```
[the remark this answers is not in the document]
dbu: oops. i assumed that `parse_url('')` would fail, but actually it gives an array with path => ''.

i add a check to fail if location is empty, as that would make no sense.

src/Plugin/RedirectPlugin.php
```diff
@@ -231,8 +231,34 @@ private function createUri(ResponseInterface $redirectResponse, RequestInterface
...
+            // this is a relative-path reference
```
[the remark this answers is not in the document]
dbu: clarifying

src/Plugin/RedirectPlugin.php
dbu: oh indeed. copy-pasted from stackoverflow :see_no_evil: 
i tried to make it a bit better

dbu:
thanks for the feedback. i updated all points you discovered

ostrolucky (approved):

PR #222, opened 2022-09-29
remove body on redirection if needed

### LINKED ISSUE #205: RedirectPlugin sends body in GET requests

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no (except when relying on the behaviour)
| Deprecations?   | no
| Related tickets | fixes #205
| Documentation   | https://github.com/php-http/documentation/pull/298
| License         | MIT

#### What's in this PR?

Remove the body and content-type and content-length headers when following a redirection.

#### Why?

Sending a body to a different target than the intended one is usually not desired and can pose a security risk.

#### Checklist

- [x] Documentation pull request created https://github.com/php-http/documentation/pull/298

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## 2.6.0 - 2022-09-29
+
+- [RedirectPlugin] Redirection of non GET/HEAD requests with a body now removes the body on follow-up requests, if the
+  HTTP method changes. To do this, the plugin needs to find a PSR-7 stream implementation. If none is found, you can
+  explicitly pass a PSR-17 StreamFactoryInterface in the `stream_factory` option.
+  To keep sending the body in all cases, set the `stream_factory` option to null explicitly.
+
 ## 2.5.1 - 2022-09-29
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


GrahamCampbell:
This should only be done if there's a change in origin.

GrahamCampbell (approved):

dbu:
thanks. good point, i fixed the changelog to be clear about when we remove the body, and also explained it in the documentation

PR #226, opened 2023-04-27
Remove dependency on php-http/message-factory

nicolas-grekas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

This package does not require `php-http/message-factory`. Since all interfaces from it have been removed, it's better to drop the dep.

PR #228, opened 2023-05-17
fix tests

dbu:
they broke in #226 because we actually relied on php-http/message-factories for tests and it was installed through indirect dependency but no longer...


GrahamCampbell (approved):

GrahamCampbell:
Hmmmm, these changes look correct, but incomplete. Not sure why yet...

dbu:
i gave up with phpspec and converted the test to phpunit. much less clutter imho :)

phpstan.neon.dist
```diff
@@ -31,6 +31,27 @@ parameters:
             count: 1
             path: src/EmulatedHttpClient.php
 
+        # we still support the obsolete RequestFactory for BC but do not require the package anymore
```
dbu, line 34: there is something with stubs one can do in phpstan but did not get it to work.

we will get rid of all this when we do a new major version anyways, so did not want to spend more time on it.

PR #231, opened 2023-07-10
Allow "guzzlehttp/psr7" v2

felliper:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     },
     "require-dev": {
         "doctrine/instantiator": "^1.1",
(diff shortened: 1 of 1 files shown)


dbu:
some tests break with the BC breaks of version 2.

the psr7 library is only a dev dependency, so it should not block installation of client-common together with psr7 v2 in an application.

what is your aim with this change?

PR #233, opened 2023-11-03
Fix static ci

dbu:


dbu:
@Ndiritu could you help me here please to figure out what annotations we need to put to avoid the phpstan errors after https://github.com/php-http/promise/pull/27?

Ndiritu:
Sure

> @Ndiritu could you help me here please to figure out what annotations we need to put to avoid the phpstan errors after [php-http/promise#27](https://github.com/php-http/promise/pull/27)?

Sure, taking a look @dbu

ste93cry:
I've been hit in one of the projects I contribute to by PHPStan complaining that

`Method handleRequest() has parameter [...] with generic interface Http\Promise\Promise but does not specify its types: T`

This has happened indirectly because the project implements a custom plugin. Since such method is expected to return a `ResponseInterface`, the solution that seems to work is the following:

```diff
/**
- * @param callable(RequestInterface): Promise $next
+ * @param callable(RequestInterface): Promise<ResponseInterface> $next
- * @param callable(RequestInterface): Promise $first
+ * @param callable(RequestInterface): Promise<ResponseInterface> $first
 *
- * @return Promise
+ * @return Promise<ResponseInterface>
 */
```

dbu:
@ste93cry we try to fix just the phpstan issue in #234 but seem to have a problem with phpstan (or some confusion on how to correctly annotate)

PR #234, opened 2023-11-07
Fix pending PHPStan errors in PR #233

### LINKED ISSUE #235: PHPStan error Promise but does not specify its types: T
**PHP version**: 8.1.22

**Description**
```
Method handleRequest() has parameter [...] with generic interface Http\Promise\Promise but does not specify its types: T
```

**How to reproduce**
```php
<?php

declare(strict_types=1);

use Http\Client\Common\Plugin;
use Http\Promise\Promise;
use Psr\Http\Message\RequestInterface;

final class Authentication implements Plugin
{
    public function handleRequest(RequestInterface $request, callable $next, callable $first): Promise
    {
        $request = $request->withHeader('Authorization', 'token bla');

        return $next($request);
    }
}
```

PHPStan config:
```neon
parameters:
    level: max
    checkMissingIterableValueType: false
    treatPhpDocTypesAsCertain: false
    paths:
        - src
```

PHPStan playground: https://phpstan.org/r/52113abc-06c9-4058-a970-4be7b8c9b311
_ignore `Http\Client\Common\Plugin` error_

**Possible Solution**
You can add a custom PHPDoc with the correct code, but it seems a bit redundant.
https://github.com/php-http/client-common/pull/233#issuecomment-1796282251 has an example of this.

```php
    /**
     * @param callable(RequestInterface): Promise<ResponseInterface> $next
     * @param callable(RequestInterface): Promise<ResponseInterface> $first
     *
     * @return Promise<ResponseInterface>
     */
    public function handleRequest(RequestInterface $request, callable $next, callable $first): Promise
```

**Additional context**
Seems https://github.com/php-http/client-common/pull/234 is already on the way to get it fixed!
But I could hardly find ANY information about "issue". So creating a new issue for documentation purposes.

/cc @ste93cry @Ndiritu

Ndiritu:
| Q               | A
| --------------- | ---
| Bug fix?        | fix #235
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Documentation   | if this is a new feature, link to pull request in https://github.com/php-http/documentation that adds relevant documentation
| License         | MIT

#### What's in this PR?

- Fixes PHPStan failures in [#233](https://github.com/php-http/client-common/pull/233) by adding appropriate generic types to Promise PHPDocs
- Removes previous PHPStan error suppressions that are resolved.

#### To Do

- [ ] pending review/release of https://github.com/php-http/promise/pull/30

### DIFF
--- a/src/Deferred.php
+++ b/src/Deferred.php
@@ -10,6 +10,8 @@ use Psr\Http\Message\ResponseInterface;
 
 /**
  * A deferred allow to return a promise which has not been resolved yet.
+ *
+ * @implements Promise<ResponseInterface>
  */
 final class Deferred implements Promise
 {
@@ -51,6 +53,11 @@ final class Deferred implements Promise
         $this->onRejectedCallbacks = [];
     }
 
+    /**
+     * @param callable(ResponseInterface): ResponseInterface|null $onFulfilled
+     * @param callable(\Throwable): ResponseInterface|null $onRejected
+     * @return Promise<ResponseInterface>
+     */
     public function then(callable $onFulfilled = null, callable $onRejected = null): Promise
     {
         $deferred = new self($this->waitCallback);
@@ -90,6 +97,7 @@ final class Deferred implements Promise
 
     /**
      * Resolve this deferred with a Response.
+     * @param ResponseInterface $response
      */
     public function resolve(ResponseInterface $response): void
     {
@@ -122,6 +130,9 @@ final class Deferred implements Promise
         }
     }
 
+    /**
+     * {@inheritDoc}
+    */
     public function wait($unwrap = true)
     {
         if (Promise::PENDING === $this->state) {
--- a/src/Deferred.php
+++ b/src/Deferred.php
@@ -51,9 +51,6 @@ final class Deferred implements Promise
         $this->onRejectedCallbacks = [];
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function then(callable $onFulfilled = null, callable $onRejected = null): Promise
     {
         $deferred = new self($this->waitCallback);
@@ -86,9 +83,6 @@ final class Deferred implements Promise
         return $deferred;
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function getState(): string
     {
         return $this->state;
@@ -128,9 +122,6 @@ final class Deferred implements Promise
         }
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function wait($unwrap = true)
     {
         if (Promise::PENDING === $this->state) {
--- a/src/HttpAsyncClientDecorator.php
+++ b/src/HttpAsyncClientDecorator.php
@@ -20,8 +20,6 @@ trait HttpAsyncClientDecorator
     protected $httpAsyncClient;
 
     /**
-     * {@inheritdoc}
-     *
      * @see HttpAsyncClient::sendAsyncRequest
      */
     public function sendAsyncRequest(RequestInterface $request)
--- a/src/HttpAsyncClientDecorator.php
+++ b/src/HttpAsyncClientDecorator.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace Http\Client\Common;
 
 use Http\Client\HttpAsyncClient;
+use Http\Promise\Promise;
+use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\RequestInterface;
 
 /**
@@ -21,6 +23,7 @@ trait HttpAsyncClientDecorator
 
     /**
      * @see HttpAsyncClient::sendAsyncRequest
+     * @return Promise<ResponseInterface>
      */
     public function sendAsyncRequest(RequestInterface $request)
     {
--- a/src/HttpAsyncClientEmulator.php
+++ b/src/HttpAsyncClientEmulator.php
@@ -17,15 +17,11 @@ use Psr\Http\Message\ResponseInterface;
 trait HttpAsyncClientEmulator
 {
     /**
-     * {@inheritdoc}
-     *
      * @see HttpClient::sendRequest
      */
     abstract public function sendRequest(RequestInterface $request): ResponseInterface;
 
     /**
-     * {@inheritdoc}
-     *
      * @see HttpAsyncClient::sendAsyncRequest
      */
     public function sendAsyncRequest(RequestInterface $request)
--- a/src/HttpAsyncClientEmulator.php
+++ b/src/HttpAsyncClientEmulator.php
@@ -23,6 +23,7 @@ trait HttpAsyncClientEmulator
 
     /**
      * @see HttpAsyncClient::sendAsyncRequest
+     * @return \Http\Promise\Promise<ResponseInterface|\Throwable>
      */
     public function sendAsyncRequest(RequestInterface $request)
     {
--- a/src/HttpClientPool/HttpClientPool.php
+++ b/src/HttpClientPool/HttpClientPool.php
@@ -52,17 +52,11 @@ abstract class HttpClientPool implements HttpClientPoolInterface
      */
     abstract protected function chooseHttpClient(): HttpClientPoolItem;
 
-    /**
-     * {@inheritdoc}
-     */
     public function sendAsyncRequest(RequestInterface $request)
     {
         return $this->chooseHttpClient()->sendAsyncRequest($request);
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function sendRequest(RequestInterface $request): ResponseInterface
     {
         return $this->chooseHttpClient()->sendRequest($request);
--- a/src/HttpClientPool/HttpClientPool.php
+++ b/src/HttpClientPool/HttpClientPool.php
@@ -7,6 +7,7 @@ namespace Http\Client\Common\HttpClientPool;
 use Http\Client\Common\Exception\HttpClientNotFoundException;
 use Http\Client\Common\HttpClientPool as HttpClientPoolInterface;
 use Http\Client\HttpAsyncClient;
+use Http\Promise\Promise;
 use Psr\Http\Client\ClientInterface;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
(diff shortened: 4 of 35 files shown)


dbu (approved):
thanks!

dbu:
lets ignore the cs errors here, they will be fixed in #233

dbu:
i tagged https://github.com/php-http/promise/releases/tag/1.2.1 and restarted the phpstan build. there are still some failures :thinking:

Ndiritu:
> i tagged https://github.com/php-http/promise/releases/tag/1.2.1 and restarted the phpstan build. there are still some failures 🤔

Taking a look.

Ndiritu:
@dbu looking into the failures related to the `$onRejected` callable.

PHPStan seems to have some limitations in recognizing that a `ClientExceptionInterface` is still a `Throwable`. Not recognizing the inheritance structure seems to only be an issue when using `callable([type]): [type]` syntax. (Looking for a relevant issue on the PHPStan repo to link here or creating one myself)

- I'm proposing that we ignore the `ClientExceptionInterface` errors in this lib.
- While testing the callable syntax I've outline the impact of expanding the parameter type for `$onRejected` to `\Throwable` for a developer who may have previously been expecting an `\Exception` and is using PHPStan. 

Please see the link below.

https://phpstan.org/r/ab8a6122-a065-4f37-8b65-b66633d212bf

dbu:
if even we can't make it work correctly, we should revert the annotations on the promise. otherwise everybody has to ignore the errors and everybody will complain first.

have you been able to check with the phpstan authors if your example should work? they are really good at improving phpstan, if it makes sense they can maybe fix it soonish.

dbu:
thanks a lot for looking into this!

that would be this interface, right? https://github.com/php-fig/http-client/blob/master/src/ClientExceptionInterface.php says it extends \Throwable. is it possible to narrow down to a more specific interface in this context?

if even we can't make it work correctly, we should revert the annotations on the promise. otherwise everybody has to ignore the errors and everybody will complain first.

have you been able to check with the phpstan authors if your example should work? they are really good at improving phpstan, if it makes sense they can maybe fix it soonish.

ste93cry:
I just tried the same example with PSalm and the same error is thrown, so either two static analysis tools have the same issue or something is wrong with the annotations 🤔

ste93cry:
I reasoned a bit on the linked playground, and I think that PHPStan is right (btw, the same error is reported by Psalm). What's happenning is that we're declaring in the interface that `$onRejected` can accept any `Throwable`. However, in the implementation, we are narrowing the argument type to be a specific exception class: that's not possible because it would mean crashing the script as soon as something is passed to it that is neither an instance of that class nor a subclass, so the tool reports it correctly.

Ndiritu:
Thank you for breaking this down @ste93cry! I get it now.
Would it be correct to say that the ideal fix then is to update this lib's `onRejected` parameter types to `\Throwable` instead of `ClientExceptionInterface`? Which would not be a breaking change if I understand correctly

Ndiritu:
Alternatively, reverting the `onRejected` PHPDoc for the `Promise` interface to `callable|null` to prevent developers having to make changes. And probably specifying `\Throwable` in a major rev of the Promise lib?

https://phpstan.org/r/4ea07ab8-ccd1-4354-8373-e14eb14231f4

src/Deferred.php
```diff
@@ -122,6 +130,9 @@ public function reject(ClientExceptionInterface $exception): void
         }
     }
 
+    /**
+     * {@inheritDoc}
+    */
```
dbu, line 135: we don't do inheritdoc when there is nothing else. if there is no docblock, most tools take the docblock of the parent/interface

src/Deferred.php
```diff
@@ -90,6 +97,7 @@ public function getState(): string
 
     /**
      * Resolve this deferred with a Response.
+     * @param ResponseInterface $response
```
dbu, line 100: we don't add phpdoc when it does not add information

```suggestion
```

src/Deferred.php
```diff
@@ -51,6 +53,11 @@ public function __construct(callable $waitCallback)
         $this->onRejectedCallbacks = [];
     }
 
+    /**
+     * @param callable(ResponseInterface): ResponseInterface|null $onFulfilled
```
dbu, line 57: this should not be necessary with the `@implements` in the class docblock, it should come automatically from the covariant in the interface.

src/HttpAsyncClientEmulator.php
```diff
@@ -23,6 +23,7 @@ abstract public function sendRequest(RequestInterface $request): ResponseInterfa
 
     /**
      * @see HttpAsyncClient::sendAsyncRequest
+     * @return \Http\Promise\Promise<ResponseInterface|\Throwable>
```
dbu, line 26: add in interface instead

src/HttpClientPool/HttpClientPool.php
```diff
@@ -52,6 +53,9 @@ public function addHttpClient($client): void
      */
     abstract protected function chooseHttpClient(): HttpClientPoolItem;
 
+    /**
+     * @return Promise<ResponseInterface>
```
dbu, line 57: not needed when HttpAsyncClient defines the return type

src/HttpClientPool/HttpClientPoolItem.php
```diff
@@ -89,6 +90,9 @@ public function sendRequest(RequestInterface $request): ResponseInterface
         return $response;
     }
 
+    /**
+     * @return Promise<ResponseInterface>
```
dbu, line 94: not needed when HttpAsyncClient defines the return type

src/HttpClientRouter.php
```diff
@@ -26,6 +27,9 @@ public function sendRequest(RequestInterface $request): ResponseInterface
         return $this->chooseHttpClient($request)->sendRequest($request);
     }
 
+    /**
+     * @return Promise<ResponseInterface>
```
dbu, line 31: not needed when HttpAsyncClient defines the return type

src/PluginClient.php
```diff
@@ -84,6 +84,9 @@ public function sendRequest(RequestInterface $request): ResponseInterface
         return $pluginChain($request)->wait();
     }
 
+    /**
+     * @return Promise<ResponseInterface>
```
dbu, line 88: not needed when HttpAsyncClient defines the return type

src/HttpAsyncClientDecorator.php
```diff
@@ -21,6 +23,7 @@ trait HttpAsyncClientDecorator
 
     /**
      * @see HttpAsyncClient::sendAsyncRequest
+     * @return Promise<ResponseInterface>
```
dbu, line 26: we need to fix this in the HttpAsyncClient instead. then it will be inferred for all the various implementations here:

https://github.com/php-http/httplug/blob/625ad742c360c8ac580fcc647a1541d29e257f67/src/HttpAsyncClient.php#L20

dbu (commented):
indeed https://github.com/php-http/promise/blob/7fa2284a91241e7fa06976369e998933d510da1b/src/Promise.php#L42C1-L42C88 is a BC break (in terms of phpdoc). we can't narrow down what type of exception the callable must accept. i think the correct thing would be to have a second covariant for the exception of the promise. then we can correctly model that the promise can only ever throw ClientExceptionInterface

can you try that?

and we should adjust the HttpAsyncClient in the httplug repository to declare what promise it returns, then we don't need to change all the various async client implementations here.

ste93cry:
>Would it be correct to say that the ideal fix then is to update this lib's `onRejected` parameter types to `\Throwable` instead of `ClientExceptionInterface`? Which would not be a breaking change if I understand correctly

Well, technically speaking it is a breaking change. You can narrow the return type in a child class because the declaration of the method is compatible, but you cannot narrow the type of an argument in a child class, because that would mean that the implementation cannot handle something that the parent declares it can. I'm not familiar with the promise library, but if `onRejected` never supported anything other than a `Throwable`, then I don't think it's wrong to just leave the typehint in the PHPDoc as it is.

>then we can correctly model that the promise can only ever throw ClientExceptionInterface

Couldn't the plugin also throw a `ServerExceptionInterface`?

ste93cry:
>Would it be correct to say that the ideal fix then is to update this lib's `onRejected` parameter types to `\Throwable` instead of `ClientExceptionInterface`? Which would not be a breaking change if I understand correctly

Well, technically speaking it is a breaking change. You can narrow the return type in a child class because the declaration of the method is compatible, but you cannot narrow the type of an argument in a child class, because that would mean that the implementation cannot handle something that the parent declares it can. For this reason, since adding the typehint would mean that people have to change their child classes to take it into account, the same is true when adding the typehint in the DocBlock and then using a static analysis tool.

>then we can correctly model that the promise can only ever throw ClientExceptionInterface

Couldn't the plugin also throw a `ServerExceptionInterface`?

dbu:
there is no ServerExceptionInterface in [PSR-18](https://www.php-fig.org/psr/psr-18/), the client exception is short to mean "http client exception" and psr18 says that every exception thrown must be a client exception. the async client follows the same logic in https://github.com/php-http/httplug/blob/2.x/src/HttpAsyncClient.php (except that for BC it has https://github.com/php-http/httplug/blob/2.x/src/Exception.php) 

imo the correct solution is to make the exception customizable in the promise as well and in async client customize the promise to use `Http\Client\Exception` as the phpdoc text already explains but could not technically specify.

ste93cry:
I wonder whether it makes sense to insist on trying to typehint `onRejected` more than now: the static analysis tool cannot infer which exception is thrown from it, so it cannot validate anyway that the next promise in the chain will receive a `ClientExceptionInterface`.

dbu:
i think for documentation purposes its good to be clear what kind of exception may be thrown / has to be handled.

ste93cry:
I came up with [this](https://phpstan.org/r/5d453cfa-c57f-4668-b0e2-d8d801c44a45) solution that seems to work fine 🎉Of course, the `$onRejected` typehint is not checked, but PHPStan does not complain at least. I think it's the best we can do with the current tools.

dbu:
I played around in the phpstan playground and came up with this: https://phpstan.org/r/6ec99fad-fd67-4f64-ac0d-ba135e207700

The upside is that the async client can declare what type of exception the onRejected callback must expect, to match with the current text description in phpdoc and the implementations of adapters.

any thoughts on that? i plan to implement this soonish if no objections come up.

PR #232, opened 2023-11-02
Allow using Symfony 7 dependencies

javiereguiluz:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

Similar to https://github.com/php-http/curl-client/pull/82. This is needed to use this package in Symfony 7 applications.

PR #237, opened 2024-01-21
chore: add testrun for symfony 7

Chris53897:
Please see https://github.com/php-http/client-common/pull/232#issuecomment-1833494282

Support for Symfony 7 is added in "phpspec/phpspec" 7.5.0
https://github.com/phpspec/phpspec/releases/tag/7.5.0

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -75,6 +75,8 @@ jobs:
             php-version: '7.4'
           - symfony: '6'
             php-version: '8.2'
+          - symfony: '7'
+            php-version: '8.2'
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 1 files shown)


dbu:
something is not working with those "LTS" builds - https://github.com/php-http/client-common/actions/runs/7600541313/job/20722422472?pr=237 is the one that says symfony 7, but it installs symfony 3 things :scream: 

we should probably set up flex with shivammatur like they do here https://github.com/doctrine/DoctrineBundle/commit/42f5834825eaa967af833722c336d5dd05cbd4f3 and make sure the major version is enforced.

dbu:
the problem with the current setup: we install flex at the same time as updating the dependencies, so flex is not yet present and does not enforce versions.

Chris53897:
@dbu
I think `composer global` was the missing part. But maybe you are right too.
Because i did remove more parts, because i think they are obsolete.
Please have a look if i did it correct.

I changed Symfony LTS according to the real LTS.
https://symfony.com/releases

dbu (approved):
ah, `global` is a solution as well.

and we missed to set the `SYMFONY_REQUIRE` variable? thanks for cleaning this up!

PR #239, opened 2024-01-21
chore: bump dev dependencies

Chris53897:
use ProphecyTrait

I need to check if ci-pipeline is green.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,12 +21,13 @@
         "symfony/polyfill-php80": "^1.17"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


dbu:
okay, we found a good reason to drop the php 7 versions. (that will also allow to reduce the range of allowed require-dev dependencies, we no longer will need 3 major versions of phpunit and prophecy...)

do you want to drop the 7 versions in a seprate MR (and add a changelog entry about it)? and then pick up this again?

Chris53897:
Yes i will drop php 7 in a seperate PR

PR #241, opened 2024-01-21
chore: bump github actions "OskarStark/phpstan-ga" version 0.x => 1.13.0

Chris53897:


.github/workflows/static.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         run: composer remove --dev friends-of-phpspec/phpspec-code-coverage phpspec/phpspec
 
       - name: PHPStan
-        uses: OskarStark/phpstan-ga@0.12.32
+        uses: OskarStark/phpstan-ga@1.13.0
```
dbu: install is failing.

in other repos, we do not specify the version. that is a bit dangerous too, as new versions can "break the build" with new reports. can we require 1.* ? (not sure about the syntax for wildcard)

Chris53897: I did change it to the recommended integration (without a version number)
https://github.com/OskarStark/phpstan-ga

dbu (approved):
thanks!

Issue #242, opened 2024-01-21
Drop support for PHP < 7.4?


Chris53897:
Combined usage for PHP 7.1, to 7.3 is at 4 Percent for Version 2.
https://packagist.org/packages/php-http/client-common/php-stats#2

xabbuh:
Would dropping support enable other things? For example, could the code be simplified because some code that is only present because of the support for PHP 7.3 could be removed?

dbu:
yeah not much value in supporting it anymore. on the other hand, i usually only drop old versions when they hinder anything. if we would move to 8.0 or 8.1, we could modernize the code. i don't think dropping 7.1-7.3 allows to modernize anything.

if you want to work on it, i suggest we move to 8.1 directly (people on legacy php can keep using the legacy versions of this library. it is pretty stable, so no big hassle to maintain). and modernize the code where we can with new language features and strict types where possible without BC breaks.

eventually we could do a new major version with BC breaks, to have strict typing everywhere.

Chris53897:
Sounds good. I will work on this. I will set PHP 8.1 as minimum in that case.

dbu:
thanks a lot! and as we just saw in  #239 the legacy support does become cumbersome.

please for now no BC breaks, so that we can have a 2.x version with the cleanups and symfony 7 support.

jtojnar:
`composer install` will now fail as guzzlehttp/psr7 1 is EOL and full of security issues:

```
No composer.lock file present. Updating dependencies to latest instead of installing from lock file. See https://getcomposer.org/install for more information.
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires guzzlehttp/psr7 ^1.4, found guzzlehttp/psr7[1.4.0, ..., 1.9.1] but these were not loaded, because they are affected by security advisories ("PKSA-vznr-tgp9-fd7d", "PKSA-7qs6-zvnz-h66r", "PKSA-gm5x-j3mz-71n9", "PKSA-jj5t-2zs1-dcfm", "PKSA-hn62-zkx4-1y5q", "PKSA-gvzg-s447-b5b5"). Go to https://packagist.org/security-advisories/ to find advisory details. To ignore the advisories, add their IDs to the "policy.advisories.ignore-id" config or add the package to "policy.advisories.ignore". To turn the feature off entirely, you can set "policy.advisories.block" to false.
```

But we cannot just bump it since it does not support PHP 7.1.

PR #243, opened 2024-01-22
feat: drop support for php < 8.1, drop support for symfony < 5.4 (WIP)

Chris53897:
https://github.com/php-http/client-common/issues/242

Just a quick view and open for disussion what to do next.
Needs rebasing after the other 2 PRs are merged.

#### Checklist

- [] Documentation pull request created (if not simply a bugfix)

#### To Do

- [ ] to Discuss

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -15,9 +15,6 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
-      - name: Remove phpspec
-        run: composer remove --dev friends-of-phpspec/phpspec-code-coverage phpspec/phpspec
-
       - name: PHPStan
         uses: docker://oskarstark/phpstan-ga
         env:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,8 +2,8 @@ name: tests
 
 on:
   push:
-      branches:
-          - '*.x'
+    branches:
+      - '*.x'
   pull_request:
 
 jobs:
@@ -36,7 +36,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php: ['7.1', '7.4', '8.2', '8.3']
+        php: ['8.1']
 
     steps:
       - name: Checkout code
@@ -61,12 +61,12 @@ jobs:
     strategy:
       matrix:
         include:
-          - symfony: '5'
+          - symfony: '5.4.*'
             php-version: '8.1'
-          - symfony: '6'
+          - symfony: '6.4.*'
             php-version: '8.2'
-          - symfony: '7'
-            php-version: '8.3'
+          - symfony: '7.0.*'
+            php-version: '8.2'
 
     steps:
       - name: Checkout code
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php: ['7.1', '7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3']
+        php: ['8.1', '8.2', '8.3']
 
     steps:
       - name: Checkout code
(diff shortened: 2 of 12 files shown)


.github/workflows/static.yml
```diff
@@ -15,11 +15,8 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
-      - name: Remove phpspec
-        run: composer remove --dev friends-of-phpspec/phpspec-code-coverage phpspec/phpspec
-
       - name: PHPStan
-        uses: OskarStark/phpstan-ga@0.12.32
+        uses: docker://oskarstark/phpstan-ga
```
dbu, line 19: can you please rebase on 2.x so we only see the changes about php 8?

.github/workflows/tests.yml
```diff
@@ -12,11 +12,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php: ['7.1', '7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3']
+        php: ['8.1', '8.2', '8.3']
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v3
```
dbu: this seems wrong

composer.json
```diff
@@ -11,22 +11,22 @@
...
+        "symfony/options-resolver": "^5.4 || ^6.0 || ^7.0"
     },
     "require-dev": {
-        "doctrine/instantiator": "^1.1",
-        "guzzlehttp/psr7": "^1.4",
-        "nyholm/psr7": "^1.2",
-        "phpspec/phpspec": "^5.1 || ^6.3 || ^7.1",
-        "phpspec/prophecy": "^1.10.2",
-        "phpunit/phpunit": "^7.5.20 || ^8.5.33 || ^9.6.7"
+        "doctrine/instantiator": "^2.0",
```
dbu: do we start to conflict with instantiator 1 or is that simply not installable with our php / symfony versions?

composer.json
```diff
@@ -11,22 +11,22 @@
...
+        "guzzlehttp/psr7": "^2.6",
```
dbu, line 23: guzzle changed from a function stream_for to a utility class with static methods. that is the reason for some of the test fails. i suggest to change those tests to create the stream with the new method.

dbu (commented):
thanks a lot!

the failing tests (apart from the stream_for issue) are probably also about missing mocking that is now detected because of strict typing of 3rd party libraries. do you look into those?

.github/workflows/static.yml
Chris53897: Yes. I did just now.

composer.json
Chris53897: I do not know. But it is a dev-dependency. So i usally do not care about older versions

dbu: do we need to depend on instantiator at all?

if it is an optional dependency of us, i worry that if we are no longer compatible with 1, people could install us with instantiator 1 and end up with broken code.

if we actually need the dependency and 1.x no longer works, we should mark < 2.0 as conflict.

.gitignore
```diff
@@ -9,3 +9,4 @@
 /vendor/
 
 .phpunit.result.cache
```
dbu: does phpunit 10 still create this file name, or can we drop this from ignores?

.github/workflows/static.yml
dbu: thanks, neat!

composer.json
Chris53897: I will look into this. Locally i just run PHPUnit Tests.
The guzzle related errors are fixed now.

.gitignore
Chris53897: I did run the migrration script. I changed it now to use the default one.
The behaviour of result cache changed some times in PHPUnit 10.

composer.json
Chris53897: I removed it. Looks like it it not needed.

dbu: thanks. 
maybe it was added at some point to prevent issues with lowest build, or something :shrug:

nunoplopes:
ping, could this be revived? It would be great to remove those unneeded dependencies for PHP 8.
@Chris53897 do you have time to rebase this?

Chris53897:
I resolved the conflicts. But there are errors. Maybe i have time at end of march to have a deeper look.

PR #250, opened 2025-11-05
Allow Symfony 8

VincentLanglet:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

Hi @dbu,
I'm adding support for SF8 

#### Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,10 +25,6 @@ jobs:
           tools: composer:v2
           coverage: none
 
-      - name: Emulate PHP 8.3
-        run: composer config platform.php 8.3.999
-        if: matrix.php == '8.4'
-
       - name: Install PHP dependencies
         run: composer update --prefer-dist --no-interaction --no-progress
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -71,8 +71,10 @@ jobs:
             php-version: '7.4'
           - symfony: '6.4.*'
             php-version: '8.2'
-          - symfony: '7.0.*'
-            php-version: '8.2'
+          - symfony: '7.4.*'
+            php-version: '8.3'
+          - symfony: '8.0.*'
+            php-version: '8.4'
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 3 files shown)


dbu:
thanks for the pull request. i triggered a build and the phpstan error is also happening on 2.x branch: https://github.com/php-http/client-common/pull/251

can you check why the LTS build fails? we can drop 4.4 build if that is problematic, but 5.4, 6.4 and 7.4 should still work. from the error it looks like something in the setup for the test is not right when 7.4 fails because 8.0

.github/workflows/tests.yml
```diff
@@ -71,8 +67,10 @@ jobs:
             php-version: '7.4'
           - symfony: '6.4.*'
             php-version: '8.2'
-          - symfony: '7.0.*'
-            php-version: '8.2'
+          - symfony: '7.4.*'
+            php-version: '8.3'
```
dbu, line 72: the build error might be related to what happens when we use php 8.3 instead of 8.2, or because we require 7.4 instead of 7.0.  7.4 has no stable release yet, do we need to allow dev versions for the 7.4 and 8.0 build?

VincentLanglet: I'm not sure why 7.4 fail ; I'll try to remove 7.4 first to see if it works on 8.0 first.

Anyway I discover it will be block by https://github.com/phpspec/phpspec

dbu: i started the build. and thanks for starting https://github.com/phpspec/phpspec/pull/1498, lets hope that can be wrapped up quickly.

the 8.0 build fails because of phpspec.

dbu:
please add `fail-fast: false` inside the LTS strategy configuration, on the same level as `matrix:`. then all builds will be run even if one of them fails.

.github/workflows/tests.yml
dbu: i converted the tests to phpunit :-|

